PR #1097, opened 2014-12-11
Search - Boost the current tenant

simong:
This is a simple addition that boosts search results from the current tenant. I suppose we'll have to play with the boost factor. This impacts searches when sharing resources with users and the general search.


simong:
Nothing to follow-up, back to @nicolaasmatthijs

nicolaasmatthijs:
Looks good. Merged.

Created a follow-up ticket at https://github.com/oaeproject/Hilary/issues/1104

PR #1106, opened 2015-01-08
Add "content body search" to library searches

simong:


nicolaasmatthijs:
@simong : Is this ready for review?

node_modules/oae-library/lib/api.search.js
```diff
@@ -81,7 +83,22 @@ module.exports.registerLibrarySearch = function(searchName, resourceTypes, optio
...
 
-                var query = SearchUtil.createQueryStringQuery(opts.q);
+                var query = {
+                    'bool': {
+                        'should': [
+                            // Search on basic metadata properties such as display name or description
+                            SearchUtil.createQueryStringQuery(opts.q),
+
+                            // Search on content comments or discussion messages
+                            SearchUtil.createHasChildQuery(ContentConstants.search.MAPPING_CONTENT_COMMENT, SearchUtil.createQueryStringQuery(opts.q, ['body']), 'max'),
```
nicolaasmatthijs: Does this mean that both content and discussion library search queries would have redundant clauses? Is that something we want to avoid?

nicolaasmatthijs:
Left a comment. Back to @simong for follow-up. If no follow-up is required, feel free to push this into master in preparation for the bug bash.

node_modules/oae-library/lib/api.search.js
simong: Yes they would have redundant clauses. I'd need to look into it a little more to see how we could avoid that

simong:
Pushed a follow-up commit that addresses the raised issue

nicolaasmatthijs:
Looks good. Merged.

PR #1099, opened 2015-01-15
Introduce editor role for collaborative documents

stuartf:
A new role, `editor`, should be defined for collaborative documents only. People with the `editor` role should be able to edit the Etherpad document, but shouldn't be able to do any other management of the doc (delete, etc.).

### DIFF
--- a/node_modules/oae-content/lib/api.js
+++ b/node_modules/oae-content/lib/api.js
@@ -762,33 +762,27 @@ var handlePublish = module.exports.handlePublish = function(data, callback) {
  */
 var joinCollabDoc = module.exports.joinCollabDoc = function(ctx, contentId, callback) {
     // Check if we have access to this piece of content.
-    canManage(ctx, contentId, function(err, canManage, contentObj) {
+    canEdit(ctx, contentId, function(err, canEdit, contentObj) {
         if (err) {
             return callback(err);
+        } else if (!canEdit) {
+            return callback({'code': 401, 'msg': 'You need to be a manager or editor of this piece of content to be able to join it'});
+        } else if (contentObj.resourceSubType !== 'collabdoc') {
+            return callback({'code': 400, 'msg': 'This is not a collaborative document'});
         }
 
-        canEdit(ctx, contentId, function(err, canEdit, contentObj) {
+        // Join the pad
+        Etherpad.joinPad(ctx, contentObj, function(err, data) {
             if (err) {
                 return callback(err);
-            } else if (!canManage && !canEdit) {
-                return callback({'code': 401, 'msg': 'You need to be a manager or editor of this piece of content to be able to join it'});
-            } else if (contentObj.resourceSubType !== 'collabdoc') {
-                return callback({'code': 400, 'msg': 'This is not a collaborative document'});
             }
 
-            // Join the pad
-            Etherpad.joinPad(ctx, contentObj, function(err, data) {
+            ContentDAO.Etherpad.saveAuthorId(data.author.authorID, ctx.user().id, function(err) {
                 if (err) {
                     return callback(err);
                 }
 
-                ContentDAO.Etherpad.saveAuthorId(data.author.authorID, ctx.user().id, function(err) {
-                    if (err) {
-                        return callback(err);
-                    }
-
-                    return callback(null, {'url': data.url});
-                });
+                return callback(null, {'url': data.url});
             });
         });
     });
--- a/node_modules/oae-content/lib/api.js
+++ b/node_modules/oae-content/lib/api.js
@@ -409,7 +409,7 @@ var _createFile = function(ctx, displayName, description, visibility, file, addi
  * @param  {String}         displayName             The display name of the collaborative document
  * @param  {String}         [description]           A longer description for the collaborative document
  * @param  {String}         [visibility]            The visibility of the collaborative document. One of `public`, `loggedin`, `private`
- * @param  {Object}         [additionalMembers]     Object where the keys represent principal ids that need to be added to the content upon creation and the values represent the role that principal will have. Possible values are "viewer" and "manager"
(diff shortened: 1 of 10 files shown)


stuartf:
Ready for review

node_modules/oae-content/lib/authz.js
```diff
@@ -39,6 +39,25 @@ var canManageContent = module.exports.canManageContent = function(ctx, content,
...
+ * @param  {Boolean}        callback.canManage      Whether or not the user can edit the content item
+ */
+var canEditContent = module.exports.canEditContent = function(ctx, content, callback) {
+    if (!ctx.user()) {
+        return callback(null, false);
+    } else if (ctx.user().isAdmin(content.tenant.alias)) {
+        return callback(null, true);
+    }
+
+    return AuthzAPI.hasRole(ctx.user().id, content.id, ContentConstants.roles.EDITOR, callback);
```
simong: Can we do AuthzAPI.getAllRoles and return `true` if the user is either an editor or manager?

node_modules/oae-content/tests/test-content.js
```diff
@@ -2963,15 +2963,22 @@ describe('Content', function() {
...
+                                            permissions[contexts['simon'].user.id] = 'manager';
+                                            RestAPI.Content.updateMembers(contexts['nicolaas'].restContext, contentObj.id, permissions, function(err) {
                                                 assert.ok(!err);
-                                                assert.equal(updatedContentObj.displayName, 'Sweet stuff');
-                                                callback();
+                                                RestAPI.Content.updateContent(contexts['simon'].restContext, contentObj.id, {'displayName': 'Sweet stuff'}, function(err, updatedContentObj) {
+                                                    assert.ok(!err);
+                                                    assert.equal(updatedContentObj.displayName, 'Sweet stuff');
+                                                    callback();
+                                                });
```
simong: Can you also add a few tests that editors can't make other people a manager, delete content items, etc..

simong:
Looks good, this turned out to be less code than I thought it'd be. Assigning back to @stuartf for some more tests

stuartf:
added tests

nicolaasmatthijs:
@stuartf: Can you also allow editors to share the document, even when it's private, and add a test for this? They shouldn't be able to manage the participants though (i.e. delete participants)

stuartf:
@nicolaasmatthijs so they should be able to add participants as viewer but not make any other role change?

nicolaasmatthijs:
@stuartf : That's a good question. As an editor, you would ideally be able to invite someone as a viewer or editor but not be able to add any other role changes. Not sure how well that works with what's currently in the sharing logic though.

stuartf:
It turns out we had a separate api call for sharing that is separate from explicit role management. So this allows editors to share but not manage roles including adding viewers even though that's what sharing ends up doing.

nicolaasmatthijs:
Over to @simong for review

simong:
Can you also add a test that you can't make a user an editor of a non-collabdoc?

simong:
LGTM, couple of tiny things and this can be merged in

stuartf:
There's a test for trying to make someone editor of non-collabdocs at https://github.com/stuartf/Hilary/blob/0a5c3a4adf60609dc8b0782b2d3ca76e44cb13ce/node_modules/oae-content/tests/test-content.js#L2984

stuartf:
Fixed review issues.

simong:
I've merged this into master, but only noticed after merging that we're doing another authz query to determine the editor role (for the `isEditor` flag in `getFullContentProfile`). That should probably be consolidated with the canShare logic

node_modules/oae-content/lib/api.js
```diff
@@ -939,8 +947,16 @@ var _canShareContent = function(ctx, contentObj, callback) {
             // If we can interact with the item, we can always share it
             return callback(null, true);
         } else if (contentObj.visibility === AuthzConstants.visibility.PRIVATE) {
-            // If the content is private, we need to be a manager to share it
-            return AuthzAPI.hasRole(user.id, contentObj.id, ContentConstants.roles.MANAGER, callback);
+            // If the content is private, we need to be a manager or editor to share it
+            return AuthzAPI.getAllRoles(ctx.user().id, contentObj.id, function(err, roles) {
```
mrvisser: This doesn't need to be `return`ed because it isn't sending back to the caller

node_modules/oae-content/lib/api.js
```diff
@@ -939,8 +947,16 @@ var _canShareContent = function(ctx, contentObj, callback) {
...
+                if (err) {
+                    callback(err);
```
mrvisser: This needs to be returned

node_modules/oae-content/lib/api.js
```diff
@@ -939,8 +947,16 @@ var _canShareContent = function(ctx, contentObj, callback) {
...
+                }
+                var _isEditorOrManager = _.some(roles, function(role) {
```
mrvisser: Doesn't need to be prefixed with a `_` because it isn't an internal parameter or method, it's just a regular variable.

node_modules/oae-content/lib/api.js
mrvisser: nm, narrow github scope syndrome. It does need to be returned.

node_modules/oae-content/lib/authz.js
```diff
@@ -39,6 +39,32 @@ var canManageContent = module.exports.canManageContent = function(ctx, content,
...
+var canEditContent = module.exports.canEditContent = function(ctx, content, callback) {
+    if (!ctx.user()) {
+        return callback(null, false);
+    } else if (ctx.user().isAdmin(content.tenant.alias)) {
+        return callback(null, true);
+    }
+
+    return AuthzAPI.getAllRoles(ctx.user().id, content.id, function(err, roles) {
+        if (err) {
+            callback(err);
```
mrvisser: Needs to be `return`ed

PR #1113, opened 2015-01-21
PDF preview generation issue

simong:
A preview for the PDF at https://marist.oaeproject.org/content/rr/mkKyQqwV could not be successfully created. Can we look into whether this is caused by something on our side or whether it's one of those PDFs that html2pdfex can't process?

### DIFF
--- a/node_modules/oae-preview-processor/lib/processors/file/pdf.js
+++ b/node_modules/oae-preview-processor/lib/processors/file/pdf.js
@@ -28,9 +28,6 @@ var OaeUtil = require('oae-util/lib/util');
 var PreviewConstants = require('oae-preview-processor/lib/constants');
 var PreviewUtil = require('oae-preview-processor/lib/util');
 
-var _pdftkBinary = null;
-var _pdftkTimeout = 120000;
-
 var _pdf2htmlEXBinary = null;
 var _pdf2htmlEXTimeout = 120000;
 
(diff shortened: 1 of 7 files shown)


simong:
Attached a commit that removes the pdftk dependency and just uses graphicsmagick to generate a thumbnail. Assigning to @stuartf for review

node_modules/oae-preview-processor/lib/processors/file/pdf.js
```diff
@@ -264,60 +248,24 @@ var _convertPDFToHTMLPages = function(ctx, path, pagesDir, callback) {
...
- * @param  {Object}              callback.err    An error that occurred, if any
- * @api private
- */
-var _convertToPNG = function(ctx, input, output, width, callback) {
-    var cmd = util.format('gm convert +adjoin -define pdf:use-cropbox=true -density 150 -resize %s -quality 100 "%s" "%s"', width, input, output);
-     // Execute the command
+    // Convert the first page to a png file
+    var width = PreviewConstants.SIZES.PDF.LARGE;
+    var output = pagesDir + '/page.1.png';
+    var cmd = util.format('gm convert +adjoin -define pdf:use-cropbox=true -density 150 -resize %s -quality 100 "%s[0]" "%s"', width, path, output);
```
stuartf: Should we use the `gm` node module here instead of calling gm directly?

stuartf:
Looks good except one small question about how we're calling `gm`

simong:
Used the gm module rather than spawning it directly. Back to @stuartf

PR #1107, opened 2015-01-23
Embedding links to external downloadable files

stuartf:
When a link is added that points to a file that would be downloaded (e.g. ppt, zip, etc.), the link should be marked as non-embedable to prevent a download prompt every time the content profile is opened.


node_modules/oae-preview-processor/lib/processors/link/default.js
```diff
@@ -103,6 +103,8 @@ var generatePreviews = module.exports.generatePreviews = function(ctx, contentOb
             }
         }
 
+        ctx.addPreviewMetadata('targetType', response.headers['content-type']);
```
nicolaasmatthijs: This will at least need to be confirmed through tests

nicolaasmatthijs:
If we decide we need to treat links to files that force a download in a special way, we'll need to add that here (and test it) as well.

nicolaasmatthijs:
Back to @stuartf for follow-up

stuartf:
added test

nicolaasmatthijs:
@stuartf : Wouldn't we also need to check that the external URL doesn't have the `Content-Disposition: attachment;` header before we mark it as embeddable? Even if the browser would be able to display it, it would still want to download it.

nicolaasmatthijs:
@stuartf : An outstanding comment still remains here

stuartf:
added check for content-disposition and test

node_modules/oae-preview-processor/lib/processors/link/default.js
```diff
@@ -86,6 +86,8 @@ var generatePreviews = module.exports.generatePreviews = function(ctx, contentOb
             ctx.addPreviewMetadata('embeddable', false);
         } else {
 
+            var forcedDownload = response.headers['content-disposition'] && response.headers['content-disposition'].split(';')[0] === 'attachment';
```
nicolaasmatthijs: @stuartf: Would you be able to add a line of docs to this?

node_modules/oae-preview-processor/lib/processors/link/default.js
```diff
@@ -96,13 +98,16 @@ var generatePreviews = module.exports.generatePreviews = function(ctx, contentOb
...
              */
-            if (response.headers['x-frame-options']) {
+            if (response.headers['x-frame-options'] || forcedDownload) {
                 ctx.addPreviewMetadata('embeddable', false);
             } else {
                 ctx.addPreviewMetadata('embeddable', true);
             }
         }
 
+        var contentType = response && response.headers? response.headers['content-type']: null;
```
nicolaasmatthijs: I think it would be better to put this inside of the `else` block, so we don't execute this when there has been an error

nicolaasmatthijs: Also, we are already assuming above that `response.headers` exists, so we can just do:

```
var contentType = response.headers['content-type'];
```

nicolaasmatthijs:
Added some additional review. Back to @stuartf for follow-up

stuartf:
fixed review issues

nicolaasmatthijs:
Added some more review on https://github.com/oaeproject/3akai-ux/pull/3967. Back to @stuartf

nicolaasmatthijs:
@stuartf : Some outstanding revier remains here

stuartf:
Are there any types other than `text/plain` that need to be added? I've been able to find that `webshot` uses `phantomjs` which uses `webkit` to do the actual page rendering, but the only documentation of webkit's supported formats I've been able to find said something vague like "webkit can render most common mime types".

stuartf:
OK, comparing to a blank screenshot now instead of trying to enumerate supported types. I also threw in a fix that improves the quality of thumbnails for direct links to images.

stuartf:
ready for review

sallakarppinen:
Seems to have conflicts that need resolving before this can be merged.

stuartf:
Fixed conflicts

sallakarppinen:
I still get a download prompt with [GitKraken .deb download](https://www.gitkraken.com/download/linux-deb), probably because it doesn't seem to set `Content-Disposition`.

node_modules/oae-preview-processor/lib/processors/link/default.js
```diff
@@ -86,6 +89,9 @@ var generatePreviews = module.exports.generatePreviews = function(ctx, contentOb
             ctx.addPreviewMetadata('embeddable', false);
         } else {
 
+            // If the Content-Disposition header is set to attachment the link isn't embeddable because the browser will prompt to save the target
+            var forcedDownload = response.headers['content-disposition'] && response.headers['content-disposition'].split(';')[0] === 'attachment';
```
sallakarppinen: As mentioned before I would add a check for `Content-Type: application/octet-stream` here as well.

node_modules/oae-preview-processor/lib/processors/link/default.js
```diff
@@ -96,30 +102,60 @@ var generatePreviews = module.exports.generatePreviews = function(ctx, contentOb
...
-                if (err) {
-                    log().error({'err': err, 'contentId': ctx.contentId}, 'Could not generate an image');
-                    return callback(err);
-                }
-
-                LinkProcessorUtil.generatePreviewsFromImage(ctx, path, null, callback);
-            });
+            var imgPath = ctx.baseDir + '/webshot.png';
+            // If the link target is an image just grab it instead of screenshotting it
+            if (contentType && contentType.substring(0, 6) === 'image/') {
```
sallakarppinen: Couldn't we use the `PreviewConstants.TYPES.IMAGE` here?

stuartf: Probably, I don't think that existed when I wrote this patch :smile:

sallakarppinen: ...back when dinosaurs still roamed the earth. :smile: Do you think you'll have time to have a look at this or should I take over?

sallakarppinen: Cool beans! Could you give me permission to push to this branch, please?

sallakarppinen:
Closed in favour of #1312.

PR #1117, opened 2015-01-27
Group profile should return creator's full profile instead of just id

stuartf:
Translated from 3akai-ux issue https://github.com/oaeproject/3akai-ux/pull/3968

> There was a point in time where we modified all `getFullProfile` function to include the full user profile of the creator. However, that doesn't seem to be present for groups (anymore). The back-end code should be adjusted to include the full profile (e.g. `https://github.com/oaeproject/Hilary/blob/master/node_modules/oae-content/lib/api.js#L155`) and appropriate tests should be added. 
> 
> @sathomas : Can you create a Hilary ticket for this and either have a go at it yourself or assign to @stuartf ?

Also, we should update the swagger docs for `/api/group/{groupid}`

### DIFF
--- a/node_modules/oae-principals/lib/api.group.js
+++ b/node_modules/oae-principals/lib/api.group.js
@@ -72,32 +72,39 @@ var getFullGroupProfile = module.exports.getFullGroupProfile = function(ctx, gro
             return callback(err);
         }
 
-        // Determine what roles, if any, the current user has on the group
-        _getAllRoles(ctx, groupId, function(err, roles) {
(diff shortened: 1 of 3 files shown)


stuartf:
ready for review

nicolaasmatthijs:
Over to @simong for review. @simong : Can you also verify that the `aboutgroup` widget is working as expected?

simong:
Good stuff, confirmed that this is compatible with 3akai-ux as well

PR #1119, opened 2015-02-03
Consolidate canShare and canEdit in the getFullContentProfile method

stuartf:
There's some room for optimisation here as currently the roles are being fetched twice.


stuartf:
I couldn't see a way to cleanly fetch the roles only once without a major refactor, this commit does reduce it to a single call when the user is a manager of the content though.

mrvisser:
If relevant, the pattern of `resolveEffectiveContentAccess` similar to those of Discussions and Folders could be useful here: https://github.com/oaeproject/Hilary/blob/master/node_modules/oae-folders/lib/authz.js#L271-L301

PR #1123, opened 2015-02-11
add CORS headers

stuartf:
browsers won't pass the cookies on cross origin requests, so this seems safe

### DIFF
--- a/node_modules/oae-util/lib/server.js
+++ b/node_modules/oae-util/lib/server.js
@@ -89,6 +89,13 @@ var setupServer = module.exports.setupServer = function(port, _config) {
         next();
     });
 
+    // Add CORS headers, cookies won't be passed in so all cross domain requests will be anonymous
+    app.use(function(req, res, next) {
+        res.header("Access-Control-Allow-Origin", "*");
+        res.header("Access-Control-Allow-Headers", "Origin, X-Requested-With, Content-Type, Accept");
+        next();
+    });
+
     return app;
 };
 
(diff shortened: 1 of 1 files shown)


nicolaasmatthijs:
@stuartf : It looks like there some issues with CORS support on IE9. Is that a concern for the STEM incubator work?

nicolaasmatthijs:

node_modules/oae-util/lib/server.js
```diff
@@ -89,6 +89,13 @@ var setupServer = module.exports.setupServer = function(port, _config) {
         next();
     });
 
+    // Add CORS headers, cookies won't be passed in so all cross domain requests will be anonymous
+    app.use(function(req, res, next) {
+        res.header("Access-Control-Allow-Origin", "*");
```
simong: You'll need to use `'` in order to pass the linting tests

stuartf:
lack of IE9 CORS support just means our demo won't work on IE9, unless we use a workaround like https://github.com/MoonScript/jQuery-ajaxTransport-XDomainRequest

stuartf:
fixed double quotes

node_modules/oae-util/lib/server.js
```diff
@@ -89,6 +89,13 @@ var setupServer = module.exports.setupServer = function(port, _config) {
         next();
     });
 
+    // Add CORS headers, cookies won't be passed in so all cross domain requests will be anonymous
+    app.use(function(req, res, next) {
+        res.header('Access-Control-Allow-Origin', '*');
+        res.header('Access-Control-Allow-Headers', 'Origin, X-Requested-With, Content-Type, Accept');
```
simong: Should we limit CORS to just GET requests by adding:
```
Access-Control-Allow-Methods: GET
```
?

stuartf: I don't see any reason to, since the browser won't forward cookies; even non-GET requests will be anonymous. Also, I think this gives us better support for oauth authenticated requests (i.e. you could ouath and create content without proxying through a backend server now)

simong:
Makes sense, merging.

PR #1125, opened 2015-02-13
latest juice breaks activity

stuartf:


nicolaasmatthijs:
@stuartf : Rather than freezing the version of juice, do you know how much work it would be to fix the incompatibility with latest?

stuartf:
I haven't had a chance to look at what it will take to fix it yet. I figured we could freeze the version until we can fix it in case someone tries to set up master before we get it fixed.

stuartf:
It looks like `juice.juiceContent()` got renamed to `juice.inlineContent()`

I'll open a new PR with the fix.

PR #1126, opened 2015-02-17
latest version of juice renamed an api call

stuartf:


nicolaasmatthijs:

simong:
Thanks, I followed-up on on the failing markdown test and merged this into master

PR #1130, opened 2015-03-02
Add test specifically for document publish idempotency

mrvisser:
There's a cryptic test that fails if you don't have your etherpad config text set to empty. This is a more descriptive test to fail in that case.

### DIFF
--- a/node_modules/oae-content/tests/test-collabdoc.js
+++ b/node_modules/oae-content/tests/test-collabdoc.js
@@ -489,6 +489,48 @@ describe('Collaborative documents', function() {
     });
 
     /**
+     * Test that verifies that no new revisions are created when an etherpad document is published
+     * without change
+     */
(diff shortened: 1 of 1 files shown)


node_modules/oae-content/tests/test-collabdoc.js
```diff
@@ -489,6 +489,48 @@ describe('Collaborative documents', function() {
     });
 
     /**
+     * Test that verifies that no new revisions are created when an etherpad document is published
```
nicolaasmatthijs: @mrvisser : You mentioned that your test error was caused by the default Etherpad text not being set to empty text. Should we have a test to validate that directly as well?

mrvisser: Good point, wouldn't hurt. I've opened #1133 to address this.

PR #1131, opened 2015-03-03
(Issue 1128) Use updated group timestamp to update libraries after join

mrvisser:
Addresses #1128

The unit test refactoring I did to try and reproduce this issue properly catches it happening. So regression test is baked into the general leave group tests.

### DIFF
--- a/node_modules/oae-library/lib/api.index.js
+++ b/node_modules/oae-library/lib/api.index.js
@@ -170,6 +170,14 @@ var update = module.exports.update = function(indexName, libraryVisibilities, re
         }
     };
 
+    log().trace({
+        'indexName': indexName,
+        'libraryVisibilities': libraryVisibilities,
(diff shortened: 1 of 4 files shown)


nicolaasmatthijs:

nicolaasmatthijs:
@mrvisser @simong : Does there need to be a "migration" step here to clean up any membership libraries that might have gotten out of sync? If so, we should also document as part of the upgrade guide.

mrvisser:
@nicolaasmatthijs yes it would be a good idea to nuke all group memberships libraries. We can probably just truncate the LibraryIndex CF

mrvisser:
I've added a "Next Release" section to the upgrade guide and put migration instructions there: https://github.com/oaeproject/Hilary/wiki/OAE-Upgrade-Guide#next-release

PR #1132, opened 2015-03-03
(Issue 1129) Ensure author can delete own message in a private discussion

mrvisser:
... and folder

At the time of doing the "is creator" check, we didn't have a fully accurate `canInteract` check, only implicit. We need the explicit `canInteract` check at that time, therefore we may as well just do `getEffectiveRole`.

### DIFF
--- a/node_modules/oae-discussions/lib/api.authz.js
+++ b/node_modules/oae-discussions/lib/api.authz.js
@@ -255,7 +255,7 @@ var canDeleteDiscussionMessage = module.exports.canDeleteDiscussionMessage = fun
             return callback(null, true);
         }
 
-        // It's not our message or we cannot interact and we are not manager, so we cannot delete
+        // It's not our message or we cannot interact and we are not a manager, so we cannot delete
         // this message
         return callback(null, false);
     });
--- a/node_modules/oae-discussions/lib/api.authz.js
+++ b/node_modules/oae-discussions/lib/api.authz.js
@@ -243,19 +243,21 @@ var canDeleteDiscussionMessage = module.exports.canDeleteDiscussionMessage = fun
         return callback(null, false);
     }
 
-    AuthzAPI.resolveImplicitRole(ctx, discussion.id, discussion.tenant.alias, discussion.visibility, DiscussionsConstants.roles.ALL_PRIORITY, function(err, implicitRole, canInteract) {
+    AuthzAPI.resolveEffectiveRole(ctx, discussion.id, discussion.tenant.alias, discussion.visibility, DiscussionsConstants.roles.ALL_PRIORITY, function(err, effectiveRole, canInteract) {
         if (err) {
             return callback(err);
-        } else if (implicitRole === DiscussionsConstants.roles.MANAGER) {
+        } else if (effectiveRole === DiscussionsConstants.roles.MANAGER) {
             // Managers can always delete messages
             return callback(null, true);
         } else if (canInteract && message.createdBy === user.id) {
-            // The user is still able to interact with this tenant, they can delete their own message
+            // So long as the user is still able to interact with this discussion, they can delete
+            // their own message
             return callback(null, true);
         }
 
-        // It's not our message, or we cannot implicitly interact. We need explicit manager access to delete the message
-        return AuthzAPI.hasRole(user.id, discussion.id, DiscussionsConstants.roles.MANAGER, callback);
+        // It's not our message or we cannot interact and we are not manager, so we cannot delete
+        // this message
+        return callback(null, false);
     });
 };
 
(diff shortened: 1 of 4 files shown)


mrvisser:
Note this affected both discussion and folders, content is not impacted, however Content has no `canInteract` protection at all, so it's not entirely clear if it does need it. I would guess yes, but how do we prioritize it?

node_modules/oae-discussions/lib/api.authz.js
```diff
@@ -243,19 +243,21 @@ var canDeleteDiscussionMessage = module.exports.canDeleteDiscussionMessage = fun
...
         } else if (canInteract && message.createdBy === user.id) {
-            // The user is still able to interact with this tenant, they can delete their own message
+            // So long as the user is still able to interact with this discussion, they can delete
+            // their own message
             return callback(null, true);
         }
 
-        // It's not our message, or we cannot implicitly interact. We need explicit manager access to delete the message
-        return AuthzAPI.hasRole(user.id, discussion.id, DiscussionsConstants.roles.MANAGER, callback);
+        // It's not our message or we cannot interact and we are not manager, so we cannot delete
```
nicolaasmatthijs: typo. I'll fix.

nicolaasmatthijs: Should be `a manager`

node_modules/oae-discussions/tests/test-discussions.js
```diff
@@ -2668,36 +2668,37 @@ describe('Discussions', function() {
 
             DiscussionsTestsUtil.setupMultiTenantPrivacyEntities(function(publicTenant, publicTenant1, privateTenant, privateTenant1) {
 
-                // Create message structure on the public discussion
-                RestAPI.Discussions.createMessage(publicTenant.loggedinUser.restContext, publicTenant.publicDiscussion.id, 'Message1 parent on public', null, function(err, publicMessage1) {
+                // Add a manager and member to the discussion
+                var updates = {};
```
nicolaasmatthijs: Would there be any value in running through the same test for a public, private and logged in discussion?

node_modules/oae-folders/lib/authz.js
```diff
@@ -407,19 +407,21 @@ var canDeleteFolderMessage = module.exports.canDeleteFolderMessage = function(ct
...
         } else if (canInteract && message.createdBy === user.id) {
-            // The user is still able to interact with this tenant, they can delete their own message
+            // So long as the user is still able to interact with this discussion, they can delete
+            // their own message
             return callback(null, true);
         }
 
-        // It's not our message, or we cannot implicitly interact. We need explicit manager access to delete the message
-        return AuthzAPI.hasRole(user.id, folder.groupId, FoldersConstants.roles.MANAGER, callback);
+        // It's not our message or we cannot interact and we are not manager, so we cannot delete
```
nicolaasmatthijs: `a manager`

nicolaasmatthijs:
> Content has no canInteract protection at all, so it's not entirely clear if it does need it. I would guess yes, but how do we prioritize it?

@mrvisser : What would the `canInteract` protection for content provide? When we try to delete a message on a content item, we first of all do `getContent`, which will determine if the user is allowed to see the content item. If they are, they are essentially able to interact with the content item.

nicolaasmatthijs:
Left some minor feedback and a question. Back to @mrvisser for follow-up.

mrvisser:
> Would there be any value in running through the same test for a public, private and logged in discussion?

Probably, I'll add these test routines

> What would the canInteract protection for content provide? When we try to delete a message on a content item, we first of all do getContent, which will determine if the user is allowed to see the content item. If they are, they are essentially able to interact with the content item.

That's not true. A less obvious but relevant example here is: If you are able to get to a public content item that is part of a private tenant (or a tenant not in your "network"), you can view it but cannot interact with it. In our current public+private model, this is possible when a public tenant becomes private later on.

mrvisser:
I've followed up. Back to @nicolaasmatthijs

mrvisser:
Coveralls really seem to like my work. I'm pretty sure I improved test coverage here :) @nicolaasmatthijs historically how accurate are these `0.01` coverage increase / decrease reports? Should I analyze this?

nicolaasmatthijs:
> historically how accurate are these 0.01 coverage increase / decrease reports? Should I analyze this?

We're not entirely sure how accurate these increases/decreases really are. So far, we haven't worried about 0.01/0.02 decreases in PRs, so I wouldn't worry too much about it.

nicolaasmatthijs:
Left some really minor final feedback. Back to @mrvisser for follow-up.

nicolaasmatthijs:
Actually, nvm, I'll just make the changes and will merge

nicolaasmatthijs:
Followed up and merged.

PR #1137, opened 2015-03-09
Additional review items from collabdoc editor changes

stuartf:
I've left a few late comments here:

https://github.com/oaeproject/Hilary/pull/1099

The most important thing is the un-`return`ed error handling which risks invoking the callback multiple times. There are a couple minor things that could be changed at the same time.


stuartf:
ready for review

PR #1141, opened 2015-03-13
URGENT Group Home Page Blank

simong:
I've posted this ticket in the UX area as well, sorry, wasn't sure where to put it.

I've just been alerted to an issue with OAE group ins our main Marist OAE instance.  I've been able to replicate in on my computer (Windows 8.1 FireFox 36.0.1) but have not researched beyond this....

When you go to a Group site you get the outside frame of the page but not the activity feed or navigation.  A good example is: https://marist.oaeproject.org/group/marist/lyJ02SmQp

In addition to normal concerns over such things, we are using OAE for collaboration associated with the LAK conference and folks have critical documents in Groups that we cannot access.  Thus, there is significant concern here at the moment.

I'm going to post this via GitHub as well in a moment.

Josh

### DIFF
--- a/node_modules/oae-principals/lib/api.group.js
+++ b/node_modules/oae-principals/lib/api.group.js
@@ -72,39 +72,45 @@ var getFullGroupProfile = module.exports.getFullGroupProfile = function(ctx, gro
             return callback(err);
         }
 
-        PrincipalsUtil.getPrincipal(ctx, group.createdBy, function(err, createdBy) {
+        // Determine what roles, if any, the current user has on the group
+        _getAllRoles(ctx, groupId, function(err, roles) {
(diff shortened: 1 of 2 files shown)


nicolaasmatthijs:
@JoshBaron : A fix has now been submitted and merged. We will be rolling this out to the production environment next.

nicolaasmatthijs:
@JoshBaron : The 10.1.2 release containing this fix has now been deployed to the production environment. Can you confirm that you are now able to access the group?

JoshBaron:
Yes, I can now see the group page but the 404 errors I just reported are still present.  I'm now wondering if the fix may have caused some file links to break?

PR #1143, opened 2015-03-13
Revert juice back to 0.5.0

mrvisser:


mrvisser:
Tests now seem to succeed locally. Assigning to @simong for review.

PR #1140, opened 2015-03-16
passport-google-oauth updated and is now using google+

simong:
I've fixed our version to 0.1.5 which works for now but we need to look into how we have to proceed with this.

### DIFF
--- a/node_modules/oae-authentication/config/strategies.js
+++ b/node_modules/oae-authentication/config/strategies.js
@@ -32,7 +32,7 @@ module.exports = {
             'enabled': new Fields.Bool('Google Authentication Enabled', 'Allow Google authentication for tenant', false),
             'key': new Fields.Text('Google client ID', 'Google client ID', '822036948319.apps.googleusercontent.com', {'suppress': true}),
             'secret': new Fields.Text('Google client secret', 'Google client secret', 'b5ccr5noy0j_gzj4kj42uuwm', {'suppress': true}),
-            'hostedDomain': new Fields.Text('Google hosted domain', 'Google hosted domain. e.g., Google apps domain (optional)', '')
+            'domains': new Fields.Text('Google domain(s)', 'A comma-separated lists of allowed email domains (optional)', '')
         }
     },
     'twitter': {
--- a/node_modules/oae-authentication/config/strategies.js
+++ b/node_modules/oae-authentication/config/strategies.js
@@ -32,7 +32,7 @@ module.exports = {
             'enabled': new Fields.Bool('Google Authentication Enabled', 'Allow Google authentication for tenant', false),
             'key': new Fields.Text('Google client ID', 'Google client ID', '822036948319.apps.googleusercontent.com', {'suppress': true}),
             'secret': new Fields.Text('Google client secret', 'Google client secret', 'b5ccr5noy0j_gzj4kj42uuwm', {'suppress': true}),
-            'domains': new Fields.Text('Google domain(s)', 'A comma-separated lists of allowed email domains (optional)', '')
+            'domains': new Fields.Text('Google domain(s)', 'A comma-separated list of allowed email domains (optional)', '')
         }
     },
     'twitter': {
--- a/node_modules/oae-authentication/lib/strategies/google/init.js
+++ b/node_modules/oae-authentication/lib/strategies/google/init.js
@@ -47,8 +47,8 @@ module.exports = function() {
             'clientID': key,
             'clientSecret': secret,
             'scope': [
-                'https://www.googleapis.com/auth/userinfo.profile',
-                'https://www.googleapis.com/auth/userinfo.email'
+                'profile',
+                'email'
             ],
             'callbackURL': AuthenticationUtil.constructCallbackUrl(tenant, AuthenticationConstants.providers.GOOGLE)
         },  function(accessToken, refreshToken, profile, done) {
--- a/node_modules/oae-authentication/lib/strategies/google/init.js
+++ b/node_modules/oae-authentication/lib/strategies/google/init.js
@@ -13,7 +13,9 @@
  * permissions and limitations under the License.
  */
 
+var _ = require('underscore');
 var GoogleStrategy = require('passport-google-oauth').OAuth2Strategy;
+var util = require('util');
 
 var ConfigAPI = require('oae-config');
 var Context = require('oae-context').Context;
@@ -42,6 +44,17 @@ module.exports = function() {
         // We fetch the config values *in* the getPassportStrategy so it can be re-configured at run-time.
         var key = AuthenticationConfig.getValue(tenant.alias, AuthenticationConstants.providers.GOOGLE, 'key');
         var secret = AuthenticationConfig.getValue(tenant.alias, AuthenticationConstants.providers.GOOGLE, 'secret');
+        var domains = AuthenticationConfig.getValue(tenant.alias, AuthenticationConstants.providers.GOOGLE, 'domains').split(',');
+
+        // Ensure we can do simple string comparisons by filtering empty domains and trimming out spaces
+        domains = _.chain(domains)
+            .map(function(domain) {
+                return domain.trim();
+            })
+            .filter(function(domain) {
+                return (domain && domain.length > 0);
+            })
+            .value();
 
         var passportStrategy = new GoogleStrategy({
             'clientID': key,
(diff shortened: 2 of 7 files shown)


simong:
I've attached some code that addresses this. Here are some basic steps to verify this PR:

 - On master: enable google authentication
 - On master: create a user account by signing in through google
 - On master: Create some content with that user
 - Sign out
 - Switch to this branch
 - Run npm install
 - In the Google developer's console, enable Google+ integration (I've done this for the default key that works for cam.oae.com already)
 - Restart Hilary
 - Sign back in with the user
 - Assert you signed into the correct account

simong:
Maybe also try out configured the `hostedDomain` when bugbashing/verifying

simong:
I pushed a commit that addresses #1147

node_modules/oae-authentication/config/strategies.js
```diff
@@ -32,7 +32,7 @@ module.exports = {
             'enabled': new Fields.Bool('Google Authentication Enabled', 'Allow Google authentication for tenant', false),
             'key': new Fields.Text('Google client ID', 'Google client ID', '822036948319.apps.googleusercontent.com', {'suppress': true}),
             'secret': new Fields.Text('Google client secret', 'Google client secret', 'b5ccr5noy0j_gzj4kj42uuwm', {'suppress': true}),
-            'hostedDomain': new Fields.Text('Google hosted domain', 'Google hosted domain. e.g., Google apps domain (optional)', '')
+            'domains': new Fields.Text('Google domain(s)', 'A comma-separated lists of allowed email domains (optional)', '')
```
mrvisser: Will there be any migration needed with this config key change?

node_modules/oae-authentication/lib/strategies/google/init.js
```diff
@@ -42,13 +44,24 @@ module.exports = function() {
...
+        var domains = AuthenticationConfig.getValue(tenant.alias, AuthenticationConstants.providers.GOOGLE, 'domains').split(',');
+
+        // Ensure we can do simple string comparisons by filtering empty domains and trimming out spaces
+        domains = _.chain(domains)
+            .map(function(domain) {
+                return domain.trim();
+            })
+            .filter(function(domain) {
+                return (domain && domain.length > 0);
+            })
```
mrvisser: We can just use underscore's `.compact()` instead of this filter

mrvisser: I haven't seen in what way we want to compare things yet, but should be lowercase in addition to trimming?

node_modules/oae-authentication/lib/strategies/google/init.js
```diff
@@ -56,16 +69,28 @@ module.exports = function() {
...
             var email = profile.emails[0].value;
 
-            // Re-use the email address as the externalId.
+            // Ensure the email belongs to a domain we allow
+            if (!_.isEmpty(domains)) {
+                var emailDomain = email.split('@')[1];
+                if (!_.contains(domains, emailDomain)) {
+                    var err = {
+                        'message': util.format('You tried to sign in with an that belongs to a domain (%s) that is not allowed access', emailDomain),
+                        'reason': 'domain_not_allowed'
```
mrvisser: What is this `reason` key?

node_modules/oae-authentication/config/strategies.js
simong: Yes, I made a note for that in the wiki https://github.com/oaeproject/Hilary/wiki/OAE-Upgrade-Guide#google-authentication

node_modules/oae-authentication/lib/strategies/google/init.js
simong: It's hard to pass information from an OAE strategy through passport back to OAE. This reason key will get checked by AuthenticationUtil.handlePassportError. Because there's an error, authentication will fail and the user will be redirected to: `/?authentication=failed&reason=domain_not_allowed`.

If we ever get support tickets or we want to show an appropriate error message, that reason value could be used.

node_modules/oae-authentication/lib/util.js
```diff
@@ -158,9 +158,17 @@ var logAuthenticationSuccess = module.exports.logAuthenticationSuccess = functio
 var handlePassportError = module.exports.handlePassportError = function(req, res, next) {
     return function(err) {
         if (err) {
-            if (err.message && err.message.indexOf('Unknown authentication strategy') === 0) {
+            // An OAE specific error
```
mrvisser: OAE-specific

node_modules/oae-authentication/config/strategies.js
mrvisser: `A comma-separated list of...`

node_modules/oae-authentication/lib/strategies/google/init.js
mrvisser: I see now

node_modules/oae-authentication/tests/test-external-strategies.js
```diff
@@ -1293,4 +1293,165 @@ describe('Authentication', function() {
...
+
+    describe('Google authentication', function() {
+
+        /**
+         * Configure a set of allowed domains with the `localhost` tenant's google authentication
+         *
+         * @param  {String}         domains     The comma-separated list of allowed domains
+         * @param  {Function}       callback    Standard callback function
+         */
+        var setGoogleDomains = function(domains, callback) {
```
mrvisser: Should be prefixed with a `_`

node_modules/oae-authentication/tests/test-external-strategies.js
```diff
@@ -1293,4 +1293,165 @@ describe('Authentication', function() {
...
+            configUpdate['oae-authentication/google/domains'] = domains;
+            ConfigTestUtil.updateConfigAndWait(globalAdminRestContext, global.oaeTests.tenants.localhost.alias, configUpdate, function(err) {
+                assert.ok(!err);
+            });
+
+            // Wait until the authentication api has finished refreshing its strategies
+            AuthenticationAPI.once(AuthenticationConstants.events.REFRESHED_STRATEGIES, function(tenant) {
+                return callback();
+            });
+        };
```
mrvisser: Can we wrap this up into a `oae-authentication/lib/test/util.js` file, with a method name such as:

`AuthenticationTestUtil.assertUpdateAuthConfigSucceeds`. It would then take care of both the `updateConfigAndWait`, `AuthenticationAPI.once...`.

Additionally we can add some basic assertions in the generic method such as getting the configuration point afterward and ensuring it is the value we set it to whenever it is used.

node_modules/oae-authentication/tests/test-external-strategies.js
```diff
@@ -1293,4 +1293,165 @@ describe('Authentication', function() {
...
+
+        /**
+         * Mock the requests that the Passport library would make when verifying a user signing in through Google
+         *
+         * @param  {String}     email           The user's email address
+         */
+        var mockGoogleSignInRequests = function(email) {
```
mrvisser: Prefix with `_`

node_modules/oae-authentication/tests/test-external-strategies.js
```diff
@@ -1293,4 +1293,165 @@ describe('Authentication', function() {
...
+        };
+
+        /**
+         * Assert that signing in through google with a certain email address does the right thing
+         *
+         * @param  {String}     email               The email address of the user that will sign in
+         * @param  {Boolean}    expectSuccess       Whether or not google authentication is expected to succeed
+         * @param  {Function}   callback            Standard callback function
+         */
+        var assertGoogleLogin = function(email, expectSuccess, callback) {
```
mrvisser: This would be better in an `oae-authentication/lib/test/util.js` file. Something like:

`AuthenticationTestUtil.assertGoogleAuthenticationSucceeds(email, callback)` and `AuthenticationTestUtil.assertGoogleAuthenticationFails(email, <expected http code>, callback)`

The `<expected http code>` we could swap out for something that more accurately determines the kind of failure we were expecting (e.g., `reason`).

node_modules/oae-authentication/tests/test-external-strategies.js
```diff
@@ -1293,4 +1293,165 @@ describe('Authentication', function() {
...
+            // Mock the requests the google passport strategy will make
+            mockGoogleSignInRequests(email);
```
mrvisser: If the other stuff is going into AuthenticationTestUtil, then this is probably better off there as well

node_modules/oae-authentication/tests/test-external-strategies.js
```diff
@@ -1293,4 +1293,165 @@ describe('Authentication', function() {
...
+                                        assertGoogleLogin('simon@bar.com', false, function() {
+                                            assertGoogleLogin('simon@baz.com', false, function() {
+
+                                                // Try with multiple domains
+                                                setGoogleDomains('foo.com,bar.com', function() {
+                                                    assertGoogleLogin('simon@foo.com', true, function() {
+                                                        assertGoogleLogin('simon@bar.com', true, function() {
+                                                            assertGoogleLogin('simon@baz.com', false, function() {
+
+                                                                // Try with multiple domains and trailing spaces
```
mrvisser: Might be good to then try with uppercase/lowercase mix as well

mrvisser:
@simong so is authentication using Google+ tied to a user's API key or is there some kind of organizational concept for Google+ behind the scenes where you basically tie an API key to a domain?

mrvisser:
Looks good for the most part, just some recommended refactors. Back to @simong for followup before I do some dev testing.

simong:
> @simong so is authentication using Google+ tied to a user's API key or is there some kind of organizational concept for Google+ behind the scenes where you basically tie an API key to a domain?

IIRC, you need to create a project in [Google's developer console](https://console.developers.google.com/project). Here you can:
 - Create a new OAuth client. This client has:
    - An id
    - A secret token
    - Scoped to a domain
    Example
 - Enable API integrations 
   - You need to enable the Google+ API:
    Google+ API integration

simong:
Pushed some follow-up. Back to @mrvisser for more review

mrvisser:
I'll change that `2015`. This is good to merge.

FWIW, the "old" client ID and secret seems to continue to work as well, probably until the old system is complete phased out.

simong:
Yes, the old client ID and secret shouldn't have to be changed

PR #1149, opened 2015-03-18
Group delete functionality

mrvisser:
This PR adds back-end functionality for deleting groups. Here is a list of things that are covered:

Direct functionality:

* When a group is deleted, memberships libraries no longer contain it, as well as groups that have access by virtue of access through deleted groups
* When a group is deleted, it no longer shows up in search
* When a group is restored, memberships libraries are fully restored
* When a group is restored, it returns into search
* When a group still shows up in members lists when it is deleted, there is a `deleted` property that is the millis-since-the-epoch that it was deleted
* Deleting and restoring is idempotent, so any bugs/failures while deletes and restores are processing should be simple to "reply" by invoking another delete/restore for the same group
* It is possible to search deleted items similar to general search: `/api/search/deleted`
    * Tenant admins can search only deleted items from their own tenant
    * Global admins can search deleted items from any tenant by use of `scope`, similar to `general` search

Indirect functionality:

* Ground-work was implemented for AuthzAPI to support managing "deleted ids". Further work could be done to implement a bloomfilter for deleted items so deleted checks could be done completely locally at the cost of several hashes instead of IO / Cassandra hit
* The AuthzMembershipsCache is now cached as an encoded graph instead of a list. It is possible to query full memberships and full members (indirect) by virtue of accessing a MembershipsGraph or a MembersGraph from the AuthzAPI

What's *not* implemented:

* User delete is not implemented
* Note that permissions are not erased by deleting a group. Content managers are responsible for removing groups that are deleted from being a member if they want to remove access

That's all I can think of right now. Please wait until I self-review before doing a detailed review (I'll notify), but feel free to have a high-level look.


mrvisser:
@simong this is adding a new property to the search schema. When I tested locally with my old search index, everything seemed to "just work".. As far as you know should that be expected? I can't recall if ElasticSearch got better at taking in new properties and such.

nicolaasmatthijs:
@simong : When this is ready to be merged, can you merge https://github.com/oaeproject/3akai-ux/pull/3997 as well?

mrvisser:
Assigned to @simong for review.

etc/migration/10.0-to-10.1/create_10.0.0_schema.cql3
```diff
@@ -0,0 +1,289 @@
+CREATE KEYSPACE "oaeTest" WITH replication = {'class': 'SimpleStrategy', 'replication_factor': '1'}  AND durable_writes = true;
```
simong: This folder will now have to be renamed to `10.1-to-10.2`

node_modules/oae-authz/lib/internal/graph.js
```diff
@@ -0,0 +1,179 @@
...
+ *
+ * Unless required by applicable law or agreed to in writing,
+ * software distributed under the License is distributed on an "AS IS"
+ * BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express
+ * or implied. See the License for the specific language governing
+ * permissions and limitations under the License.
+ */
+
+var _ = require('underscore');
+var Graph = require('data-structures').Graph;
```
simong: Neat!

node_modules/oae-authz/lib/internal/graph.js
```diff
@@ -0,0 +1,179 @@
...
+};
+
+/**
+ * Override the base Graph addNode function to include the `id` of the node. This will make it
+ * easier to walk the graph
+ *
+ * @param  {String}     nodeId      The id of the node
+ * @param  {Object}     [metadata]  Additional metadata to apply to the node. This is only applied if the node is new
+ * @return {Object}                 The node object that was created. `null` if a node with the given
+ *                                  id already existed
```
simong: Can the jsdocparser handle multi-line return statements?

node_modules/oae-authz/lib/internal/graph.js
```diff
@@ -0,0 +1,179 @@
...
+ * @return {Node[]}                         The array of nodes that are visited while traversing the graph
+ * @api private
+ */
+AuthzGraph.prototype._traverse = function(nodeId, getEdgeFn, nextNodeProperty, _nodes, _visitedIds) {
+    var self = this;
+
+    _nodes = _nodes || [];
+    _visitedIds = _visitedIds || {};
+
+    // Don't "Visit" this node only if it hasn't been visited already
```
simong: `// Don't "visit" this node if it has been seen already`

node_modules/oae-principals/lib/internal/dao.js
```diff
@@ -211,6 +210,86 @@ var updatePrincipal = module.exports.updatePrincipal = function(principalId, pro
...
+var deletePrincipal = module.exports.deletePrincipal = function(principalId, callback) {
+    var deleted = Date.now();
+
+    // Determine if the principal was already deleted in the authz index before we set them and flip
+    // the principals deleted flag
+    AuthzDelete.isDeleted([principalId], function(err, wasDeleted) {
+        if (err) {
+            return callback(err);
+        }
+
```
simong: Do we need to check the `wasDeleted` here?

node_modules/oae-principals/lib/internal/dao.js
```diff
@@ -211,6 +210,86 @@ var updatePrincipal = module.exports.updatePrincipal = function(principalId, pro
...
+                    // the principal as being deleted if it wasn't set before
+                    if (!wasDeleted[principalId]) {
+                        AuthzDelete.unsetDeleted(principalId);
+                    }
+
+                    return callback(err);
+                }
+
+                // Invalidate the principal from the user cache if they are a user
+                return OaeUtil.invokeIfNecessary(isUser(principalId), invalidateCachedUsers, [principalId], callback);
```
simong: We've been caught out before with `invokeIfNecessary` where the unit tests weren't hitting the other option. As we're not exposing deleteUser yet, can you add a unit test that calls `PrincipalsDAO.deletePrincipal` with a userId directly and then verifies whether the user object has been updated in the cache?

node_modules/oae-principals/lib/internal/dao.js
```diff
@@ -211,6 +210,86 @@ var updatePrincipal = module.exports.updatePrincipal = function(principalId, pro
...
+                    // the principal as being deleted if was set before
+                    if (wasDeleted[principalId]) {
+                        AuthzDelete.setDeleted(principalId);
+                    }
+
+                    return callback(err);
+                }
+
+                // Invalidate the principal from the user cache if they are a user
+                return OaeUtil.invokeIfNecessary(isUser(principalId), invalidateCachedUsers, [principalId], callback);
```
simong: Same thing here

node_modules/oae-authz/lib/api.js
```diff
@@ -490,12 +486,14 @@ var _checkGroupMembershipsForUser = function(userId, groupIds, callback) {
...
-            _explodeGroupMemberships(userId, function(err, allMemberships) {
+            _explodeGroupMemberships(userId, function(err, graph) {
                 if (err) {
                     return callback(err);
                 }
 
-                // Return the matching groups
+                // The memberships are the nodes flattened from the graph, with the subject user
+                // removed
+                var allMemberships = _.chain(graph.getNodes()).pluck('id').without(userId).value().sort();
```
simong:
```javascript
var allMemberships = _.chain(graph.getNodes())
    .pluck('id')
    .without(userId)
    .value()
    .sort();
```

Out of interest, at what number of characters do you wrap? We can probably standardize on something.

node_modules/oae-authz/lib/api.js
```diff
@@ -711,48 +872,92 @@ var _getIndirectPrincipalMembershipsByExplosion = function(principalId, start, l
 };
 
 /**
- * Given some principal, get their entire group membership ancestry. In other words, get all the groups to which they belong,
- * and the groups to all those groups belong, and so on.
+ * Create a group memberships graph for all the provided principals. The resulting graph is
+ * guaranteed to contain all principal ids, even if some or all of them don't have any memberships.
+ * The result is a graph that is possibly several disjoint graphs, as it's not necessary that all
+ * provided principals have a memberships graph that intersect with one another
```
simong: Can you add that this will return the full (so both direct and indirect) memberships?

node_modules/oae-authz/lib/api.js
```diff
@@ -545,14 +564,141 @@ var _explodeGroupMemberships = function(principalId, callback) {
...
+
+    // Get all the members of the given resource
+    var resourceId = _resourceIds.shift();
+    getAuthzMembers(resourceId, null, 10000, function(err, members) {
+        if (err) {
+            return callback(err);
+        }
+
+        // Add each member to the current members graph and associate the appropriate edges
+        _graph.addNode(resourceId);
```
simong: can you move `_graph.addNode(resourceId);` up to where you shift it out of the `resourceIds` array?

node_modules/oae-authz/lib/api.js
```diff
@@ -545,14 +564,141 @@ var _explodeGroupMemberships = function(principalId, callback) {
...
+
+        // For every entry in the memberships cache, we need to extract its relevant members and
+        // create the membership relations
+        _.each(rows, function(row) {
+            // The id of a group that the `principalId` is directly or indirectly a member
+            var groupId = row.get('groupId').value;
+
+            // This is the stored members information for the group. It is of the form:
+            // `{'memberId': 'g:oae:another-group', 'role': 'manager'}`
+            var memberRoles = JSON.parse(row.get('value').value);
```
simong: We probably should try / catch this and complain very loudly if it ever errors out

node_modules/oae-authz/lib/internal/graph.js
nicolaasmatthijs: So far, we've been putting all parameter descriptions on a single line

node_modules/oae-authz/lib/test/util.js
```diff
@@ -16,6 +16,114 @@
...
 var assert = require('assert');
 
+var AuthzAPI = require('oae-authz');
+var AuthzDelete = require('oae-authz/lib/delete');
+
+/**
+ * Delete the given resource by its id and ensure it completes successfully
+ *
+ * @param  {String}         resourceId      The id of the resource to try and delete
+ * @param  {Function}       callback        Invoked when all assertions pass
```
simong: `Standard callback function`

node_modules/oae-authz/tests/test-authzgraph.js
```diff
@@ -0,0 +1,221 @@
...
+    });
+
+    /**
+     * Test that verifies inbound and outbound traversal hits all the expected nodes in the correct
+     * order
+     */
+    it('verify inbound and outbound traversal', function(callback) {
+        var graph = new AuthzGraph();
+
+        // Create a circular graph a-i that is circular, but also links vowes from lowest to highest
```
simong: `vowels`

node_modules/oae-principals/lib/api.group.js
```diff
@@ -832,6 +844,119 @@ var updateGroup = module.exports.updateGroup = function(ctx, groupId, profileFie
...
+ * Determine whether or not the user in context can delete a given group
+ *
+ * @param  {Context}    ctx                 Standard context object containing the current user and the current tenant
+ * @param  {String}     groupId             The group to check
+ * @param  {Function}   callback            Standard callback function
+ * @param  {Object}     callback.err        An error that occurred, if any
+ * @param  {Boolean}    callback.canDelete  Whether or not the user can delete the group
+ * @param  {Group}      callback.group      The group object that was fetched from the database
+ */
+var canDeleteGroup = module.exports.canDeleteGroup = function(ctx, groupId, callback) {
```
simong: I realize it might be best to leave this to a later refactor, but I like the pattern of having these `canAction` functions in a separate `authz.js` file.

node_modules/oae-principals/lib/api.group.js
```diff
@@ -832,6 +844,119 @@ var updateGroup = module.exports.updateGroup = function(ctx, groupId, profileFie
...
+        }
+
+        // Mark the group as deleted
+        PrincipalsDAO.deletePrincipal(groupId, function(err) {
+            if (err) {
+                return callback(err);
+            }
+
+            // Notify consumers that a group has been deleted
+            PrincipalsEmitter.emit(PrincipalsConstants.events.DELETED_GROUP, ctx, group);
```
simong: Can you track this with mixpanel? See `oae-mixpanel/lib/eventlisteners/groups.js`

node_modules/oae-principals/lib/api.group.js
```diff
@@ -832,6 +844,119 @@ var updateGroup = module.exports.updateGroup = function(ctx, groupId, profileFie
...
+        }
+
+        // Unmark the group as deleted
+        PrincipalsDAO.restorePrincipal(groupId, function(err) {
+            if (err) {
+                return callback(err);
+            }
+
+            // Notify consumers that a group has been restored
+            PrincipalsEmitter.emit(PrincipalsConstants.events.RESTORED_GROUP, ctx, group);
```
simong: Can you track this with mixpanel? See `oae-mixpanel/lib/eventlisteners/groups.js`

node_modules/oae-authz/lib/internal/graph.js
mrvisser: Yea this should be single line. It's just the summary we tend to wrap

node_modules/oae-principals/lib/internal/dao.js
mrvisser: No we don't. We aren't avoiding re-running the `deletePrincipal` method if the principal was already deleted, we're only keeping track in case the query to set `deleted=true` below fails, we can undo the delete if necessary.

node_modules/oae-authz/lib/api.js
mrvisser: I wrap at 100. This looks like it even succeeds, I may have just not caught the margin line. I'll fix.

mrvisser: for cases where nesting is "getting bad" and it's not feasible to wrap, I tend to stop wrapping for the most part :/ For comments in this situation, I only wrap to ensure that the width of the comments don't exceed the width of the function calls they are describing.

node_modules/oae-authz/lib/test/util.js
mrvisser: For the `assert...` methods, that tag line hasn't been used because it's not our standard callback function that contains an error as the first parameter.

node_modules/oae-authz/lib/api.js
simong: That seems on par with what I'm trying to do.

With underscore's chaining, I prefer to have each action on a separate line though.

node_modules/oae-principals/lib/delete.js
```diff
@@ -0,0 +1,345 @@
...
+        // Add the callback function to the handlerArgs
+        var thisHandlerArgs = handlerArgs.concat(function(errs) {
+            // Decrement the delete counter to indicate we've finished processing this handler
+            deleteCounter.decr();
+
+            if (!_.isEmpty(errs)) {
+                return log().error({'principalId': principal.id, 'handlerName': name, 'errs': errs}, 'Error(s) occurred while trying to process a handler');
+            }
+
+            return log().info({'principalId': principal.id, 'handlerName': name}, 'Successfully processed handler');
```
simong: Should we drop this down to `debug` level?

node_modules/oae-principals/lib/delete.js
```diff
@@ -0,0 +1,345 @@
...
+    // Invoke all group restore handlers
+    invokeGroupRestoreHandlers(group);
+});
+
+
+/////////////////////
+// DELETE HANDLERS //
+/////////////////////
+
+/**
```
simong: Shouldn't these handlers go in `library.js` and `search.js` ?

node_modules/oae-principals/lib/delete.js
```diff
@@ -0,0 +1,345 @@
...
+ * @param  {AuthzGraph}     membershipsGraph    The graph of group memberships of the group
+ * @param  {AuthzGraph}     membersGraph        The graph of group members of the group
+ * @param  {Function}       callback            Standard callback function
+ * @param  {Object[]}       callback.errs       All errs that occurred while trying to fire the search update tasks, if any
+ * @api private
+ */
+var _handleInvalidateSearch = function(group, membershipsGraph, membersGraph, callback) {
+    // All members (direct and indirect, users and groups) of the group that was deleted need to
+    // have their memberships child search documents invalidated
+    var groupAndUserIds =
```
simong: No need to push _.chain to the next line, we usually just do:

```javascript
var groupAndUserIds = _.chain(membersGraph.getNodes())
    .pluck('id')
    .filter(AuthzUtil.isPrincipalId)
    .without(group.id)
    .partition(AuthzUtil.isGroupId)
    .value();
```

node_modules/oae-principals/lib/internal/dao.js
```diff
@@ -428,8 +515,14 @@ var getGroupFromRow = function(row) {
...
  * @api private
  */
-var getUserFromRow = function(row) {
-    return hashToUser(Cassandra.rowToHash(row));
+var _getUserFromRow = function(row) {
+    var hash = Cassandra.rowToHash(row);
+
+    // Helenus returns the timestamp data-type as a string. Lets convert them to millis since the
+    // epoch
+    _timestampToMillis(hash, 'deleted');
```
simong: Do users not have the `created` timestamp?

simong: Never mind, it looks like they don't

node_modules/oae-principals/lib/library.js
```diff
@@ -74,3 +85,86 @@ LibraryAPI.Search.registerLibrarySearch('memberships-library', ['group'], {
...
+            // Delete all group nodes from the graph that have been deleted
+            _.chain(deleted)
+                .keys()
+                .each(function(deletedGroupId) {
+                    graph.removeNode(deletedGroupId);
+                });
+
+            // The resulting membership will be all group nodes that are reachable VIA a full
+            // outbound edge traversal ("member of") starting from the `principalId`
+            var memberships =
```
simong: _.chain can go on this line

node_modules/oae-principals/lib/library.js
```diff
@@ -74,3 +85,86 @@ LibraryAPI.Search.registerLibrarySearch('memberships-library', ['group'], {
...
+                _.chain(graph.traverseOut(principalId))
+                    .pluck('id')
+                    .without(principalId)
+                    .value();
+
+            // Return the final memberships list, ordered in string id order
```
simong: Does the order even matter here?

mrvisser: It doesn't, and I don't think it even will even be guaranteed to be anymore in this case. When I first did this, I was under the impression that the `pageResources` thing would still page through things, then I realized I could just return all of them and force a `null` nextToken, so ordering wasn't really necessary anymore after I already wrote some of this

simong:
Excellent work as usual. Back to @mrvisser

node_modules/oae-principals/lib/internal/dao.js
mrvisser: We don't honour the `deleted` flag for users (`user.model.js` doesn't accept it). I think I should actually remove this and have the DAO give an error if a user is attempted to be deleted until we have actual support for it.

mrvisser:
Followed up, back to @simong

simong:
LGTM. Setting as On hold until after the bug bash

mrvisser:
@simong I've added some functionality to suppress the `profilePath`/`oae:profilePath` property if the entity is deleted. I'm going to flip the labels on this so the last commit can be reviewed.

nicolaasmatthijs:
@mrvisser : Is there still a way for global/tenant admins to find "deleted" groups? Ideally, they would come back in searches for global/tenant admins, so they can go to the profile page for the deleted group and restore it from there. I don't think this needs to hold up a release, but it would be nice to have from a support point of view.

mrvisser:
@nicolaasmatthijs no you can't go to the profile page of a deleted group, not even an admin. It returns a 404 and we removed the `profilePath` link so no link is even available similar to private users.

There is a "deleted" search available at `/api/search/deleted` that will search only deleted items, and a restore endpoint available at `POST /api/group/<groupId>/restore`. Those 2 things together would allow us to work in group restore functionality in the admin panel, but we've out-scoped that work.

We could work the stuff into the regular UI, it's just that the logic in the regular codepaths will be more complex to handle these scenarios and we'll have to take a different approach than simply hiding the `profilePath` from the model.

PR #1151, opened 2015-03-20
Fix tile line-height in iphone emails; Align visibility icon to bottom

mrvisser:
Addresses the line-height portion of https://github.com/oaeproject/3akai-ux/issues/3990

This was only visible on iPhone, but it seems using a px line-height works quite a bit better in the Apple Mail app.

I've also moved the visibility icons to the bottom right of the metadata rather than top right. This is consistent with newer designs.

Note that it's tough to see pixel-perfectness with a local build since the icons don't link properly, so I'd like to see this on QA0 after merged and iterate if necessary.

### DIFF
--- a/node_modules/oae-activity/emailTemplates/mail.html.jst
+++ b/node_modules/oae-activity/emailTemplates/mail.html.jst
@@ -152,9 +152,9 @@
         str += '            <tr>';
         str += '                <td valign="bottom">';
         str += '                    <div class="preview-tile-metadata">';
-        str += '                        <table cellpadding="0" cellspacing="0">';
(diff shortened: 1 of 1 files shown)


nicolaasmatthijs:
screen shot 2015-03-19 at 21 11 14

- The item title and type are still fairly close to each other. I would add a margin-bottom of 3/4px to the title of the item
- The type is 1/2px too far to the top

nicolaasmatthijs:
Left some additional visual feedback. Back to @mrvisser for follow-up

mrvisser:
Brought the resource type down a couple pixels. Back to @nicolaasmatthijs

nicolaasmatthijs:
Looks good. Merged

PR #1153, opened 2015-03-22
Made `email` a required field when creating a user

simong:
A mapping between an email address and a principal has been added which allows
us to resolve a user by their email address. Email verification is NOT
implemented yet.

This commit also moves `email` out of the options field when creating users 
and makes it a required parameter. Lots of the tests that were using 
createUser directly have been replaced with `TestsUtil.generateTestUsers`.

### DIFF
--- a/node_modules/oae-authentication/lib/api.js
+++ b/node_modules/oae-authentication/lib/api.js
@@ -163,9 +163,9 @@ var localUsernameExists = module.exports.localUsernameExists = function(ctx, ten
  * @param  {String}     username                The unique username for the global administrator
  * @param  {String}     password                The password for the global administrator
  * @param  {String}     displayName             The display name for the global administrator
- * @param  {String}     email                   The email address for the global administrator
  * @param  {String}     [opts]                  Optional user profile parameters
  * @param  {String}     [opts.locale]           The locale for the global administrator
+ * @param  {String}     [opts.email]            The email address for the global administrator
  * @param  {string}     [opts.emailPreference]  The email preference for the global administrator
  * @param  {String}     [opts.publicAlias]      The name to show when the global administrator is inaccessible to a user
  * @param  {Function}   callback                Standard callback function
@@ -174,7 +174,7 @@ var localUsernameExists = module.exports.localUsernameExists = function(ctx, ten
  * @param  {String}     callback.loginId        The *flattened* loginId for this user
  * @param  {Boolean}    callback.created        `true` if a global administrator was created with the provided username, `false` otherwise
  */
-var getOrCreateGlobalAdminUser = module.exports.getOrCreateGlobalAdminUser = function(ctx, username, password, displayName, email, opts, callback) {
+var getOrCreateGlobalAdminUser = module.exports.getOrCreateGlobalAdminUser = function(ctx, username, password, displayName, opts, callback) {
     opts = opts || {};
 
     var validator = new Validator();
@@ -184,8 +184,6 @@ var getOrCreateGlobalAdminUser = module.exports.getOrCreateGlobalAdminUser = fun
     validator.check(password, {'code': 400, 'msg': 'You must provide a password'}).notEmpty();
     validator.check(displayName, {'code': 400, 'msg': 'You must provide a display name'}).notEmpty();
     validator.check(displayName, {'code': 400, 'msg': 'A display name can be at most 1000 characters long'}).isShortString();
-    validator.check(email, {'code': 400, 'msg': 'You must provide an email address'}).notEmpty();
-    validator.check(email, {'code': 400, 'msg': 'A valid email address must be provided'}).isEmail();
     if (validator.hasErrors()) {
         return callback(validator.getFirstError());
     }
@@ -194,7 +192,7 @@ var getOrCreateGlobalAdminUser = module.exports.getOrCreateGlobalAdminUser = fun
     opts.visibility = AuthzConstants.visibility.PRIVATE;
 
     var providerProperties = {'password': password};
-    getOrCreateUser(ctx, AuthenticationConstants.providers.LOCAL, username, providerProperties, displayName, email, opts, function(err, user, loginId, created) {
+    getOrCreateUser(ctx, AuthenticationConstants.providers.LOCAL, username, providerProperties, displayName, opts, function(err, user, loginId, created) {
         if (err) {
             return callback(err);
         } else if (!created) {
@@ -229,28 +227,24 @@ var getOrCreateGlobalAdminUser = module.exports.getOrCreateGlobalAdminUser = fun
  * @param  {String}       authProvider              The authentication provider of the login id
  * @param  {String}       externalId                The desired externalId/username for this user
  * @param  {String}       displayName               The display name for the user
- * @param  {String}       email                     The email address for the user
  * @param  {String}       [opts]                    Optional user profile parameters
  * @param  {String}       [opts.locale]             The locale for the user
  * @param  {String}       [opts.email]              The email address for the user
  * @param  {String}       [opts.emailPreference]    The email preference for the user
  * @param  {String}       [opts.visibility]         The visibility of the user. One of: @see AuthzConstants.visibility
  * @param  {String}       [opts.publicAlias]        The name to show when the user is inaccessible to a user
- * @param  {String}       [opts.emailPreference]    The email preference for the user. One of: @see PrincipalsConstants.emailPreference
  * @param  {Function}     callback                  Standard callback function
  * @param  {Object}       callback.err              An error that occurred, if any
  * @param  {User}         callback.user             The user that was fetched or created
  * @param  {String}       callback.loginId          The *flattened* loginId for this user
  * @param  {Boolean}      callback.created          `true` if the user was created, `false` otherwise
  */
-var getOrCreateUser = module.exports.getOrCreateUser = function(ctx, authProvider, externalId, providerProperties, displayName, email, opts, callback) {
+var getOrCreateUser = module.exports.getOrCreateUser = function(ctx, authProvider, externalId, providerProperties, displayName, opts, callback) {
     opts = opts || {};
 
     var validator = new Validator();
     validator.check(displayName, {'code': 400, 'msg': 'You must provide a display name'}).notEmpty();
     validator.check(displayName, {'code': 400, 'msg': 'A display name can be at most 1000 characters long'}).isShortString();
-    validator.check(email, {'code': 400, 'msg': 'You must provide an email address'}).notEmpty();
-    validator.check(email, {'code': 400, 'msg': 'A valid email address must be provided'}).isEmail();
     if (validator.hasErrors()) {
         return callback(validator.getFirstError());
     }
@@ -262,7 +256,7 @@ var getOrCreateUser = module.exports.getOrCreateUser = function(ctx, authProvide
         return callback(validator.getFirstError());
     }
 
-    _getOrCreateUser(ctx, loginId, displayName, email, opts, callback);
+    _getOrCreateUser(ctx, loginId, displayName, opts, callback);
 };
 
 /**
@@ -271,7 +265,6 @@ var getOrCreateUser = module.exports.getOrCreateUser = function(ctx, authProvide
  * @param  {Context}    ctx                 Standard context object containing the current user and the current tenant
  * @param  {LoginId}    loginId             The login id to use to fetch or create the user
  * @param  {String}     displayName         The display name for the user
- * @param  {String}     email               The email address for the user
  * @param  {Object}     [opts]              Optional user profile parameters
  * @param  {Function}   callback            Standard callback function
  * @param  {Object}     callback.err        An error that occurred, if any
@@ -280,7 +273,7 @@ var getOrCreateUser = module.exports.getOrCreateUser = function(ctx, authProvide
  * @param  {Boolean}    callback.created    `true` if the user was created, `false` otherwise
  * @api private
  */
-var _getOrCreateUser = function(ctx, loginId, displayName, email, opts, callback) {
+var _getOrCreateUser = function(ctx, loginId, displayName, opts, callback) {
     opts = opts || {};
 
     _getUserIdFromLoginId(loginId, function(err, userId) {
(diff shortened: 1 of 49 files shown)


simong:
In an attempt to make the inviting guests work a more phased approach I'll try to submit PRs in the following manner:
 - Make `email` a required user property (this PR)
 - Adding email verification when signing up / updating the email address
 - When no email is provided during SSO, intercept user creation and present a form to the user
 - Invite a user by an email address

I can re-use lots of code from the other PR, but it was getting a little bit too much to do in one go. Hopefully this works out better.

simong:
I should note that this PR is not against master but to a special `inviting-guest` branch as their will (obviously) be lots of TODOs between some of the PRs.

mrvisser:
@simong can you summarize what the final impact of this PR is now that email is no longer being made a required parameter?

node_modules/oae-authentication/lib/api.js
```diff
@@ -448,7 +450,7 @@ var _createUser = function(ctx, loginId, displayName, opts, callback) {
             PrincipalsAPI.createUser(ctx, loginId.tenantAlias, displayName, opts, function(err, user) {
                 if (err) {
                     Locking.release(lockKey, lockToken, function() {
-                        callback(err);
```
mrvisser: lullllz

node_modules/oae-authentication/tests/test-auth-local.js
```diff
@@ -1173,55 +1205,75 @@ describe('Authentication', function() {
...
-                                    assert.ok(!err);
-                                    assert.ok(me.anon);
+                                    // Ensure the global admin still cannot be authenticated
+                                    var testGlobalUserRestContext = TestsUtil.createGlobalRestContext();
 
-                                    // Sanity check the user can be created through the REST endpoints
-                                    RestAPI.User.createGlobalAdminUser(globalAdminRestContext, username, username, username, null, function(err, user) {
-                                        assert.ok(!err);
-                                        assert.ok(user);
+                                    // Ensure that the credentials do not authenticate a value global administrator
```
mrvisser: Should this be `Ensure that the credentials do not authenticate a valid global administrator` ?

node_modules/oae-authentication/tests/test-auth-local.js
```diff
@@ -1173,55 +1205,75 @@ describe('Authentication', function() {
...
+                                            assert.ok(me.anon);
 
-                                            return callback();
+                                            // Sanity check the user can be created through the REST endpoints
+                                            RestAPI.User.createGlobalAdminUser(globalAdminRestContext, username, username, username, email, {}, function(err, user) {
+                                                assert.ok(!err);
+                                                assert.ok(user);
+
+                                                // Ensure when we try and create one with the same loginId, we get a 400 error
+                                                RestAPI.User.createGlobalAdminUser(globalAdminRestContext, username, username, username, email, {}, function(err, secondUser) {
```
mrvisser: So, to be clear, is this something we all agree on, to not let new accounts be created that share email addresses with other accounts?

It seems to restrict something that we've treated as a valid use of the system, which is the same person having multiple accounts on the system in different tenants.

node_modules/oae-principals/lib/api.user.js
```diff
@@ -130,52 +133,70 @@ var createUser = module.exports.createUser = function(ctx, tenantAlias, displayN
...
-        'visibility': opts.visibility,
-        'email': opts.email,
-        'emailPreference': opts.emailPreference,
-        'locale': opts.locale,
-        'publicAlias': opts.publicAlias,
-        'smallPictureUri': opts.smallPictureUri,
-        'mediumPictureUri': opts.mediumPictureUri,
-        'largePictureUri': opts.lagePictureUri
-    };
+    // Verify this email address, if any, is not already mapped to another user
```
mrvisser: Can you just drop a `TODO` in here for this nagging to happen. It's hard to predict how much time will be in between this and a fully featured nag workflow.

node_modules/oae-principals/lib/api.user.js
```diff
@@ -130,52 +133,70 @@ var createUser = module.exports.createUser = function(ctx, tenantAlias, displayN
...
+        var user = new User(tenantAlias, id, displayName, opts.email, {
+            'visibility': opts.visibility,
+            'locale': opts.locale,
+            'publicAlias': opts.publicAlias,
+            'emailPreference': opts.emailPreference,
+            'smallPictureUri': opts.smallPictureUri,
+            'mediumPictureUri': opts.mediumPictureUri,
+            'largePictureUri': opts.lagePictureUri,
+            'acceptedTC': 0,
+            'guest': opts.guest
```
mrvisser: Should guest be removed in this PR?

node_modules/oae-principals/lib/api.user.js
```diff
@@ -354,9 +375,6 @@ var importUsers = module.exports.importUsers = function(ctx, tenantAlias, userCS
                             if (!user.publicAlias || user.publicAlias === externalId) {
                                 update['publicAlias'] = displayName;
                             }
-                            if (!user.email) {
-                                update['email'] = opts.email;
-                            }
```
mrvisser: Why wouldn't we do this anymore. i.e., update a user's email address with what is in the update CSV if they don't currently have one?

node_modules/oae-principals/lib/api.user.js
```diff
@@ -475,21 +494,52 @@ var updateUser = module.exports.updateUser = function(ctx, userId, profileFields
...
+            if (err && err.code !== 404) {
                 return callback(err);
+            } else if (user && user.id !== oldUser.id) {
+                return callback({'code': 400, 'msg': 'This email address is already associated with another user'});
             }
 
-            var newUser = PrincipalsUtil.createUpdatedUser(oldUser, profileFields);
-            PrincipalsEmitter.emit(PrincipalsConstants.events.UPDATED_USER, ctx, newUser, oldUser);
-            return getUser(ctx, userId, callback);
+            // TODO: Email verification if the email address was changed
```
mrvisser: Should this not be part of this PR?

node_modules/oae-principals/lib/internal/dao.js
```diff
@@ -185,28 +216,71 @@ var updatePrincipal = module.exports.updatePrincipal = function(principalId, pro
...
+ * @param  {String}     callback.oldEmail                   The user's old email address
+ * @api private
+ */
+var _isEmailAddressUpdate = function(principalId, profileFields, callback) {
+    if (!isUser(principalId) || !profileFields.email) {
+        return callback(null, false);
     }
 
-    Cassandra.runQuery(q.query, q.parameters, function(err) {
+    getPrincipal(principalId, function(err, user) {
```
mrvisser: Before I get much further and forget, I just thought I'd mention that I haven't seen any lower-case/upper-case handling yet. emails should probably be lower-cased as early as possible when entering the API layer, if not already.

node_modules/oae-principals/lib/rest.user.js
```diff
@@ -268,14 +268,20 @@ OAE.tenantRouter.on('post', '/api/user/create', function(req, res) {
...
+        return res.status(400).send('A valid email address is required when creating a local account');
+    }
+
     /*!
      * Create a local user account
      */
     var createUser = function() {
-        var loginId = new LoginId(tenant.alias, AuthenticationConstants.providers.LOCAL, req.body.username, { password: req.body.password });
-        AuthenticationAPI.createUser(ctx, loginId, req.body.displayName, opts, function(err, newUser) {
+        AuthenticationAPI.getOrCreateUser(ctx, AuthenticationConstants.providers.LOCAL, req.body.username, { password: req.body.password }, req.body.displayName, opts, function(err, newUser, loginId, created) {
```
mrvisser: `'password'` and no spaces between the `{` / `}`'s

mrvisser:
I've gone through the non-test files and some of the test files, I've left a few comments. I will continue tomorrow on test files, hoping `w=1` makes it a little more bearable :)

node_modules/oae-authentication/tests/test-auth-local.js
nicolaasmatthijs: In principle, it's more desirable if one person can have multiple accounts in different tenants, especially in order to satisfy some of the Fold / STEM incubator use cases we currently have. Having said that, if we want a way in which we can share something with a person through an email address, I don't see a way around the restriction as we would have no idea which account to match that with.

nicolaasmatthijs: I guess we could say that you can only invite new users by email address and have to select the person you want from the autosuggest if they already exist. Would that be a viable approach?

mrvisser: I think it might so long as it doesn't force us to break some kind of privacy boundary. For example I type in an email address, now I get a list of accounts (private, loggedin, public) that are set up with that email address.

You also have to take into consideration private tenants. Not in the case that you're sharing something from a private tenant (I'm assuming that would not be allowed), but instead you've entered an email address that exists in the system but it's for a user who exists in a private tenant elsewhere. I guess you would just treat it as though there is no user with that email address in the system?

node_modules/oae-principals/lib/api.user.js
simong: We do, this is an oversight. It's already been addressed in the verification PR

simong:
@mrvisser The goal of this PR is mainly to get the refactor to RestAPI.User.createUser out of the way.

It's a first step towards ensuring users have a unique email address. Email verification is NOT part of this PR, but one of the following (I'm just waiting for this to get out of the way)

FWIW, This won't be going into *master*, but in a separate *inviting-guests* branch as it's not really production ready.

mrvisser:
@simong Ah I see, thanks for the clarification.

mrvisser:
In the `verify an email address cannot be re-used`, we'll probably want to include tests for variations in email casing.

mrvisser:
Review complete. @simong , I'm assuming you plan to create an inviting-guests branch on the `oaeproject` clone (if not already) to submit your PRs to? That way we can have the refactors out of the picture when reviewing the next one.

simong:
@mrvisser, yes this PR is already against such a branch

mrvisser:
derrrr :goat:

simong:
Rebased against `inviting-guests` (which = `master` for now)

node_modules/oae-principals/lib/api.user.js
```diff
@@ -130,52 +133,69 @@ var createUser = module.exports.createUser = function(ctx, tenantAlias, displayN
...
-        'smallPictureUri': opts.smallPictureUri,
-        'mediumPictureUri': opts.mediumPictureUri,
-        'largePictureUri': opts.lagePictureUri
-    };
+    // Verify this email address, if any, is not already mapped to another user
+    _getUserByEmailIfNecessary(opts.email, null, function(err, existingUser) {
+        if (err && err.code !== 404) {
+            return callback(err);
+        } else if (existingUser) {
+            return callback({'code': 400, 'msg': 'This email address is already associated with another user'});
```
mrvisser: We aren't supposed to be restricting this anymore. You can create as many accounts as you like with the same email.

node_modules/oae-principals/lib/internal/dao.js
```diff
@@ -184,28 +216,71 @@ var updatePrincipal = module.exports.updatePrincipal = function(principalId, pro
...
-        if (isUser(principalId)) {
-            // Update the user in cache
-
-            // Ensure the `principalId` is not part of the hash to avoid revalidating a stale cache entry
-            profileFields = _.extend({}, profileFields);
-            delete profileFields.principalId;
-            return _updateCachedUser(principalId, profileFields, callback);
-        } else {
-            return callback();
+        if (user.email !== profileFields.email) {
```
mrvisser: Not a huge deal, but this might be easier to read in an `else if` since it's a simple one-liner break like the error callback

node_modules/oae-principals/lib/internal/dao.js
```diff
@@ -347,6 +422,28 @@ var setAdmin = module.exports.setAdmin = function(adminType, isAdmin, userId, ca
...
+        if (err) {
+            return callback(err);
+        } else if (_.isEmpty(rows)) {
+            return callback({'code': 404, 'msg': 'No user found by that email'});
+        }
+
+        var principalId = rows[0].get('principalId').value;
+        return getPrincipal(principalId, callback);
+    });
+};
```
mrvisser: I'm wondering if this should be `getUsersByEmail` and return an array containing each.

node_modules/oae-principals/lib/rest.user.js
```diff
@@ -268,14 +268,20 @@ OAE.tenantRouter.on('post', '/api/user/create', function(req, res) {
     var user = ctx.user();
     var opts = _getOptionalProfileParameters(req.body);
 
+    // We enforce an email address, the API will take care of validating the actual value
+    if (!opts.email) {
+        return res.status(400).send('A valid email address is required when creating a local account');
```
mrvisser: I think we generally do `res.send(400, '....');`

node_modules/oae-principals/tests/data/users-with-password-alias.csv
```diff
@@ -0,0 +1,25 @@
...
+user-mIAUwkeNS,password3,Visser,Branden,branden@alias.test.com
+user-HlKKreaDP,password4,Gaeremynck,,simon@alias.test.com
+user-IrewPDSAw,password5,Freeman,Stuart D.,stuart@alias.test.com
+user-JjuyZssJQ,password6,Reynolds,Malcom,mal@alias.test.com
+user-HgRNdDzQd,password7,Washburn,Zoe,zoe@alias.test.com
+user-DGdlFUNzv,password8,Washburn,Hoban,wash@alias.test.com
+user-gPvjYjVHh,password9,Serra,Inara,inara@alias.test.com
+user-TUolZGzsm,password10,Cobb,Jayne,jayne@alias.test.com
+user-CIwpSRwXk,password11,Frye,Kaylee,kaylee@alias.test.com
+user-SUMbWTWKD,password13,Tam,River,river@alias.test.com
```
mrvisser: Most apartment buildings skip the 13th floor, this is the first superstition I've seen for the number 12, though :)

mrvisser:
Left a couple comments. I think the biggest thing I don't quite understand is why we're validating email is unique, might just be an oversight?

node_modules/oae-principals/lib/internal/dao.js
simong: That's what I did in the follow-up PR.. This PR only addresses the issues you've raised before we decided to take another approach. Maybe I should just close this one, and submit it as one PR?

mrvisser: That's fine, I'll just review it in the follow-up.

mrvisser:
Also, there aren't a ton of comments left here, so you can feel free to merge this into `inviting-guests` if you want and take care of it in the next PR.

PR #1144, opened 2015-03-24
Prepend protocol to new links

stuartf:
New links that don't have a protocol should be prepended with `http://` to ensure they can always be correctly embedded.

### DIFF
--- a/node_modules/oae-content/lib/api.js
+++ b/node_modules/oae-content/lib/api.js
@@ -234,6 +234,13 @@ var createLink = module.exports.createLink = function(ctx, displayName, descript
 
     var contentId = _generateContentId(ctx.tenant().alias);
     var revisionId = _generateRevisionId(contentId);
+
+    // Make sure the URL starts with a protocol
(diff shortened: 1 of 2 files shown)


stuartf:
ready for review

node_modules/oae-content/lib/api.js
```diff
@@ -234,6 +234,13 @@ var createLink = module.exports.createLink = function(ctx, displayName, descript
 
     var contentId = _generateContentId(ctx.tenant().alias);
     var revisionId = _generateRevisionId(contentId);
+
+    // Make sure the URL starts with a protocol
+    var hasProtocol = link.indexOf('://') !== -1;
```
nicolaasmatthijs: Not a huge fan of this code style. Would you mind changing this to:

```
if (link.indexOf('://') === -1) {
  link = 'http://' + link;
}
```

nicolaasmatthijs: As this is the only comment, I'll follow up myself.

nicolaasmatthijs:
Followed up and merged

PR #1155, opened 2015-03-25
(GroupProfiles) Implement members library for groups ranked on profile pic and role

mrvisser:


mrvisser:
As you can see, this is based on principals delete, so hold up for that to be merged :)

mrvisser:
I have rebased this with master, should be good for a high-level review but I still have yet to do a self-review.

nicolaasmatthijs:
@mrvisser : I wanted to do a functional review of the group profiles, but it looks like `oae-emitter` might be missing from this PR.

mrvisser:
@nicolaasmatthijs argh, didn't update the `.gitignore` etc etc... I've added it in now.

mrvisser:
Self-review is complete. Assigning to @simong for review

node_modules/oae-authz/lib/api.js
```diff
@@ -987,6 +988,42 @@ var _getInvalidateMembershipsCacheQueries = function(userIds) {
         .value();
 };
 
+/**
+ * Get all roles available for the given principal on any resource of type `resourceType`. This is
+ * a version of `getRolesForPrincipalAndResourceType` that pages.
```
simong: `... that fetches all the pages.` ?

simong:
Any reason these emitters weren't changed?

>
node_modules/oae-activity/lib/internal/emitter.js:var events = require('events');
node_modules/oae-activity/lib/internal/notifications.js:var events = require('events');
node_modules/oae-library/lib/api.index.js:var events = require('events');
node_modules/oae-principals/lib/api.user.js:var events = require('events');
node_modules/oae-util/lib/oae.js:var events = require('events');
node_modules/oae-util/lib/test/mq-util.js:var events = require('events');

mrvisser:
> Any reason these emitters weren't changed?

No reason (except for the `oae-emitter`, for obvious reasons). What the heck did I grep for to miss all those?

mrvisser:
> Any reason these emitters weren't changed?

Turns out all except 1 of these didn't actually use the events import. So I must have grepped for `.on`s or something.

mrvisser:
Followed up on all review up to this point.

node_modules/oae-content/lib/api.js
```diff
@@ -1294,7 +1294,13 @@ var removeContentFromLibrary = module.exports.removeContentFromLibrary = functio
                         return callback(err);
                     }
 
-                    LibraryAPI.Index.remove(ContentConstants.library.CONTENT_LIBRARY_INDEX_NAME, [ libraryOwnerId ], content.id, content.lastModified, function(err) {
+                    var entry = {
```
simong: Would it make sense to have a `LibraryEntry` class?

That way we could do:
```javascript
// When adding some members to a content item
var insertEntries = _.map(memberIds, function(memberId) {
    return new LibraryEntry(memberId, contentObj.lastModified, contentObj);
});
LibraryAPI.Index.insert(ContentConstants.library.CONTENT_LIBRARY_INDEX_NAME, insertEntries, callback);

// Or when removing a member
var entry = new LibraryEntry(libraryOwnerId, content.lastModified, content);
LibraryAPI.Index.remove(ContentConstants.library.CONTENT_LIBRARY_INDEX_NAME, [entry], callback);
```

mrvisser: Sure.

node_modules/oae-principals/lib/api.group.js
```diff
@@ -208,39 +217,60 @@ var getGroupMembers = module.exports.getGroupMembers = function(ctx, groupId, st
...
+ * @param  {User[]|Group[]}     callback.members        An array of the direct members of the group
+ * @param  {String}             callback.nextToken      The value to provide in the `start` parameter to get the next set of results
+ * @api private
+ */
+var _getMembersLibrary = function(ctx, group, hasRole, start, limit, callback) {
+    // Ensure proper permissions if we haven't determined that they have an explicit role
+    OaeUtil.invokeIfNecessary(!hasRole, LibraryAPI.Authz.resolveTargetLibraryAccess, ctx, group.id, group, function(err, hasAccess, visibility) {
+        if (err) {
+            return callback(err);
+        } else if (hasRole) {
```
simong: Would `resolveTargetLibraryAccess` not do this for you?

node_modules/oae-principals/lib/libraries/members.js
```diff
@@ -0,0 +1,293 @@
...
+
+    // Simply rank managers higher
+    if (role === PrincipalsConstants.roles.MANAGER) {
+        roleRank = '1';
+    }
+
+    // Picture visibility (which implies link visibility) is the most important, secondary
+    // importance given to their link visibility. Finally, all else being equal, we prefer manager
+    // profiles over members
+    return util.format('%s%s%s', pictureVisibilityRank, linkVisibilityRank, roleRank);
```
simong: Wouldn't this trip up the `_splitRankedResourceId` in `oae-library/lib/api.index` ?

node_modules/oae-principals/lib/libraries/members.js
```diff
@@ -0,0 +1,293 @@
...
+ * indicate their rank or visibility will need to change
+ *
+ * @param  {RestContext}    ctx                 The REST context of the user that performed the update
+ * @param  {User|Group}     updatedPrincipal    The updated copy of the principal
+ * @param  {User|Group}     oldPrincipal        The old copy of the principal before update
+ */
+var _updatePrincipalInMembersLibraries = function(ctx, updatedPrincipal, oldPrincipal) {
+    // If there is no change in state of their profile picture or visibility, we don't have to make
+    // any updates
+    if (!_xor(_hasProfilePicture(updatedPrincipal), _hasProfilePicture(oldPrincipal)) &&
```
simong: Isn't this just:
```
if ((_hasProfilePicture(updatedPrincipal) === _hasProfilePicture(oldPrincipal)) &&
    oldPrincipal.visibility === updatedPrincipal.visibility) {
```

?

node_modules/oae-principals/lib/libraries/members.js
```diff
@@ -0,0 +1,293 @@
...
+        oldPrincipal.visibility === updatedPrincipal.visibility) {
+        return;
+    }
+
+    // If the user updated their profile picture, we need to recalculate their rank for every group
+    // members library to which they belong
+    AuthzAPI.getAllRolesForPrincipalAndResourceType(updatedPrincipal.id, AuthzConstants.resourceTypes.GROUP, function(err, allRoles) {
+        if (err) {
+            // We should log that an error happened, but we shouldn't fail the user update process
```
simong: This comment seems out of place

node_modules/oae-principals/lib/libraries/members.js
```diff
@@ -0,0 +1,293 @@
...
+        }
+
+        return callback();
+    });
+});
+
+/*!
+ * When someone leaves a group, we should remove them from the group's members library. We don't
+ * have to do this with a synchronous `when` handler because the user generally won't notice within
+ * such small latency if the members library hasn't been updated
```
simong: Wouldn't you be able to make the same argument for joining a group? Maybe they both need to be `when` ?

node_modules/oae-principals/lib/test/util.js
```diff
@@ -995,10 +1138,24 @@ var getAllGroupMembers = module.exports.getAllGroupMembers = function(restCtx, g
...
-        return getAllGroupMembers(restCtx, groupId, opts, callback, _nextToken, _members, _responses);
+        return assertGetAllMembersLibrarySucceeds(restCtx, groupId, opts, callback, _nextToken, _members, _responses);
     });
 };
+
+/**
+ * Get a stream to some default picture
+ *
+ * @param  {String}     [path]  A path to a picture to use. Defaults to the file located at: "oae-principals/tests/data/restroom.jpg"
+ * @return {Stream}     A stream to jpg image
```
simong: We usually align the return description with the parameter descriptions

simong:
PR looks good. Couple of minor nitpickings. Back to @mrvisser

node_modules/oae-content/lib/api.js
mrvisser: Actually there are different variations of the same entity. E.g., the difference between an entry that is provided for insert (`rank`) and an entry that is provided for update (`oldRank`/`newRank`). So it's probably best we just keep it as a simple object with specific properties based on the library index method.

node_modules/oae-principals/lib/api.group.js
mrvisser: It does, but `hasRole` is much cheaper than `getRole` which `resolveTargetLibraryAccess` does. It's just a bit of a performance-related short-circuit to pass in the `hasRole`, because there is one context where we have already determined that the user has a role, which gives us enough information to proceed without getting their actual role.

node_modules/oae-principals/lib/libraries/members.js
mrvisser: If it contained a `#` then yes, but I don't see how a rank of the form `000` would trip it up

node_modules/oae-principals/lib/libraries/members.js
mrvisser: Not really, because `_hasProfilePicture` is truthy/falsey, not explicit boolean values. I should change that though rather than have `_xor`

node_modules/oae-principals/lib/libraries/members.js
mrvisser: I don't think so. When joining a group, I would think looking at the members list is something they may do next. If they don't see themselves there, then they could think something is wrong.

Contrarily, if a user leaves a group, I don't imagine they'd care to go look at the members library afterward.

That's a lot of hand-waving, though, but if anyone is really passionate about it I could switch it to be synchronous.

mrvisser:
I've either followed up or replied to all review up to this point. Back to @simong

node_modules/oae-principals/lib/libraries/members.js
simong: Oops, I totally thought it contained `#` characters

simong:
LGTM. Just waiting on the UI counter-part before merging

PR #1158, opened 2015-03-26
Search indexers can return a set of errors AND documents

simong:
An entire batch shouldn’t be aborted because 1 resource failed

### DIFF
--- a/node_modules/oae-content/lib/search.js
+++ b/node_modules/oae-content/lib/search.js
@@ -166,76 +166,77 @@ ContentAPI.on(ContentConstants.events.DELETED_COMMENT, function(ctx, comment, co
 /**
  * Produce the necessary content comment search documents.
  *
- * @param  {Object[]}   resources           The resources representing the content item whose comments to index
- * @param  {Function}   callback            Standard callback function
- * @param  {Object}     callback.err        An error that occurred, if any
- * @param  {Object[]}   callback.documents  The content comment search documents
+ * @see SearchAPI#registerChildSearchDocument
  * @api private
  */
-var _produceContentCommentDocuments = function(resources, callback, _documents) {
+var _produceContentCommentDocuments = function(resources, callback, _documents, _errors) {
     _documents = _documents || [];
+    _errors = _errors || [];
     if (_.isEmpty(resources)) {
-        return callback(null, _documents);
+        return callback(_errors, _documents);
     }
 
     var resource = resources.pop();
     if (resource.comments) {
         var documents = MessageBoxSearch.createMessageSearchDocuments(ContentConstants.search.MAPPING_CONTENT_COMMENT, resource.id, resource.comments);
         _documents = _.union(_documents, documents);
-        return _produceContentCommentDocuments(resources, callback, _documents);
+        return _produceContentCommentDocuments(resources, callback, _documents, _errors);
     }
 
     // If there were no messages stored on the resource object, we go ahead and index all comments for the content item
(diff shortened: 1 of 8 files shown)


simong:
Assigning to @mrvisser

node_modules/oae-content/lib/search.js
```diff
@@ -166,76 +166,77 @@ ContentAPI.on(ContentConstants.events.DELETED_COMMENT, function(ctx, comment, co
...
     var resource = resources.pop();
     if (resource.comments) {
         var documents = MessageBoxSearch.createMessageSearchDocuments(ContentConstants.search.MAPPING_CONTENT_COMMENT, resource.id, resource.comments);
         _documents = _.union(_documents, documents);
-        return _produceContentCommentDocuments(resources, callback, _documents);
+        return _produceContentCommentDocuments(resources, callback, _documents, _errors);
     }
 
     // If there were no messages stored on the resource object, we go ahead and index all comments for the content item
-    MessageBoxSearch.createAllMessageSearchDocuments(ContentConstants.search.MAPPING_CONTENT_COMMENT, resource.id, resource.id, function(err, documents) {
```
mrvisser: Can we use `errs` instead of `errors`?

node_modules/oae-content/lib/search.js
```diff
@@ -166,76 +166,77 @@ ContentAPI.on(ContentConstants.events.DELETED_COMMENT, function(ctx, comment, co
...
- * @param  {Function}   callback            Standard callback function
- * @param  {Object}     callback.err        An error that occurred, if any
- * @param  {Object[]}   callback.documents  The content comment search documents
+ * @see SearchAPI#registerChildSearchDocument
  * @api private
  */
-var _produceContentCommentDocuments = function(resources, callback, _documents) {
+var _produceContentCommentDocuments = function(resources, callback, _documents, _errors) {
     _documents = _documents || [];
+    _errors = _errors || [];
```
mrvisser: We should try and keep this as being falsey until an error actually happens. Doing `_.union(_errs, errs)` should still work if `_errs` is undefined so that shouldn't have to change (can you quickly verify that in the REPL? I'm pretty sure I did that for group delete stuff)

node_modules/oae-content/lib/search.js
```diff
@@ -166,76 +166,77 @@ ContentAPI.on(ContentConstants.events.DELETED_COMMENT, function(ctx, comment, co
...
 
-        var tenantAlias = revision.previewsId.split(':')[1];
+            // Move on to the next resource
+            return _produceContentBodyDocuments(resources, callback, _documents, _errors);
+        }
 
         ContentDAO.Previews.getContentPreview(revision.previewsId, 'plain.txt', function(err, preview) {
             if (err) {
-                return callback(err);
+                _errors.push(err);
```
mrvisser: This could be `_errs = _.union(_errs, [err]);` if we got the route of leaving `_errs` as undefined until an error happens

node_modules/oae-content/lib/search.js
```diff
@@ -166,76 +166,77 @@ ContentAPI.on(ContentConstants.events.DELETED_COMMENT, function(ctx, comment, co
...
+                return _produceContentBodyDocuments(resources, callback, _documents, _errors);
             }
 
+            var tenantAlias = revision.previewsId.split(':')[1];
```
mrvisser: `AuthzUtil` has a method to do this

simong:
Pushed some follow-up, also re-enabled Travis

mrvisser:
Looks good

Issue #1148, opened 2015-03-17
Swagger 2.0


simong:
This is a pretty minor issue but Swagger's 2.0 spec has been out for a while and it offers some nice things. One thing in particular would be really useful, which is the operation descriptions. It allows you to provide more documentation about an operation and go into more detail. It allows for markdown as well which is kinda nice.

Something like:
```
/**
 * @REST deleteContentContentId
 *
 * Delete a content item
 *
 * Delete a piece of content from the system. The content item will be removed from all the content libraries and/or folders it's part of automatically.
 *
 * @Server      tenant
 * @Method      DELETE
 * @Path        /content/{contentId}
etc..
 */
```

I don't think this needs any changes to the `restjsdoc` module, just some reformatting/restructring in `oae-docs`.

stuartf:
So, the second description is separate from the first one though? And if it can contain markdown I imagine that means it can have line breaks? I think this is doable, though it will take a small change in `restjsdoc`.

simong:
Yes. Although that's just a proposal, I'm more than happy to do it in other (easier?) ways. Swagger now allows you to have a short summary (~100 characters) and a larger blob of notes. Something through which we can add those notes would be really useful

stuartf:
OK, I think we might actually be able to do this without touching `restjsdoc` because everything between the `@REST` and the next `@<tag>` just goes into a description field, so we can split that on the first newline to populate the 2 different descriptions.

stuartf:
Looking at the swagger 2.0 spec it doesn't appear that there are actually 2 description fields, they just changed the `operations` array to a `paths` object, though I think some things like the markdown processing of the description field are new. So we'll need to update the version of swagger-ui on the front end, as well as re-arranging the swagger endpoint's output to match the 2.0 spec.

PR #1161, opened 2015-03-31
"My" search

mrvisser:
This should support the "My Stuff".

There still needs to be some UI stuff associated to this, which is in development. So it's possible this doesn't turn out to meet 100% of the UI requirements. However, here it is for initial review if desired.

Built on top of GroupProfiles, of course. So that will need to be merged first.

### DIFF
--- a/node_modules/oae-authz/tests/test-groups.js
+++ b/node_modules/oae-authz/tests/test-groups.js
@@ -864,7 +864,7 @@ describe('Authz Groups', function() {
          * @param  {String[]}   callback.groups         An array of group ids representing the indirect groups to which the user belongs
          * @throws {Error}                              An assertion error is thrown if an error occurs or the retrieved groups don't match the expected groups
          */
-        var assertAllIndirectMembershipsEquals = function(principalId, expectedGroupIds, callback) {
+        var _assertAllIndirectMembershipsEquals = function(principalId, expectedGroupIds, callback) {
             AuthzAPI.getAllIndirectPrincipalMemberships(principalId, function(err, groupIds) {
                 assert.ok(!err);
                 assert.deepEqual(groupIds.sort(), expectedGroupIds.slice().sort());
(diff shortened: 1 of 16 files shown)


mrvisser:
Rebased with master, self-reviewed. Assigning to @simong for review.

node_modules/oae-util/tests/test-globals.js
```diff
@@ -0,0 +1,74 @@
...
+describe('Globals', function() {
+
+    describe('Underscore', function() {
+
+        describe('oaeObj', function() {
+
+            /**
+             * Test that verifies that an object is created from a variety of input
+             */
+            it('verify an object is created from a variety of input', function() {
```
simong: I personally am fine with not adding `callback` in the arguments and not doing `return callback();` for synchronous tests, but I believe we settled on doing so

node_modules/oae-authz/tests/test-groups.js
```diff
@@ -15,10 +15,12 @@
 
 var _ = require('underscore');
 var assert = require('assert');
+var shortid = require('shortid');
```
simong: `ShortId`

simong:
At this point, I think we need a place to communicate to REST clients what `scope`s, `resourceType`s, etc.. are. Once https://github.com/oaeproject/Hilary/issues/1148 gets addressed we will be able to expose a pre-amble that goes into more detail about each endpoint. I think it'd be useful to document each search "type" that we register through `registerSearch` with Swagger separately. ie, have separate docs for:
 * /api/search/general
 * /api/search/deleted
 * /api/search/following
 * /api/search/followers
 * /api/search/{library-name}

WDYT?

mrvisser:
Yea, I agree. It would be great to be able "register" your own docs for it. Maybe it's actually better for now if we don't have the central doc, but instead have individual doc blocks for each search type that gets registered?

mrvisser:
Hopefully swagger has a way to document an enumerable string, that way we can have a semantic way of documenting the different resource types, scopes, etc...

simong:
Only had some superminor feedback, back to @mrvisser

node_modules/oae-authz/tests/test-groups.js
mrvisser: Why does this one get camel cased while all the other external deps are lowercased? I'd suggest just renaming the others to `shortid` so the style is more predictable

mrvisser:
Assigning to @simong for feedback on `shortid` dependency

node_modules/oae-util/tests/test-globals.js
```diff
@@ -0,0 +1,76 @@
...
+                    '0': 0,
+                    'null': null,
+                    'anotherNull': null,
+                    'another0': 0
+                };
+
+                // Ensure the result is the expected object, and that `source` gets overridden
+                assert.deepEqual(_.oaeExtendDefined(source, extendWith), expectedDest);
+                assert.deepEqual(source, expectedDest);
+                callback();
```
simong: You'll have to put `callback` in `it`'s parameter list

simong: You'll also have to `return` them

mrvisser: > You'll also have to `return` them

I thought we didn't `return` them for "simple methods" that didn't have async calls / error handling cases ?

node_modules/oae-authz/tests/test-groups.js
simong: I'm not sure. I think we started using PascalCasing all variables requiring modules of our own (e.g.,  PrincipalsDAO, ContentAPI, ..) but kept short single word variables such as `fs`, `assert`, `util`, ... lowercase? Either way it's all terribly inconsistent. I'd rather we keep using `ShortId` everywhere for now or settle on style and start changing all of them.

mrvisser: > I'm not sure. I think we started using PascalCasing all variables requiring modules of our own (e.g., PrincipalsDAO, ContentAPI, ..) but kept short single word variables such as fs, assert, util, ... lowercase?

I think this is what we did as well. Can we settle on it? I'll rename all `ShortId`s to `shortid`, it's not too much.

mrvisser:
I've defined the callback. Oops. Also rebased w/ master. Back to @simong .

Note this should simply go on hold as the UI portion has not gone under any review yet and it is fairly significant.

simong:
LGTM, marking as "On hold"

nicolaasmatthijs:
@simong : If this is still ready for merging, feel free to pull the trigger

simong:
@nicolaasmatthijs This was On Hold pending the UI work. I'm happy to merge it in once the UI's ready

PR #1163, opened 2015-04-09
support for permissions changes in role hierarchy

stuartf:
this is needed for https://github.com/oaeproject/3akai-ux/pull/3970

### DIFF
--- a/node_modules/oae-content/lib/api.js
+++ b/node_modules/oae-content/lib/api.js
@@ -1079,48 +1079,75 @@ var setContentPermissions = module.exports.setContentPermissions = function(ctx,
             return callback(err);
         }
 
-        // Check if the current user can manage the content
-        ContentAuthz.canManageContent(ctx, contentObj, function(err, canManage) {
+        // Check that the user isn't trying to assign any role higher than his own
+        AuthzAPI.resolveEffectiveRole(ctx, contentObj.id, contentObj.tenant.alias, contentObj.visibility, ContentConstants.roles.ALL_PRIORITY, function(err, effectiveRole, canInteract) {
             if (err) {
                 return callback(err);
-            } else if (!canManage) {
-                return callback({'code': 401, 'msg': 'You are not allowed to manage this piece of content'});
+            } else if (!canInteract) {
+                return callback({'code': 401, 'msg': 'You are not allowed to interact with this piece of content'});
             }
 
-            // Ensure that we're not crossing tenant boundaries or violating principal visibility settings
-            _checkNewContentPermissions(ctx, contentObj, permissionChanges, function(err, membershipAfterChanges, newMemberIds, updatedMemberIds, removedMemberIds, newMembers) {
+            ContentAuthz.canManageContent(ctx, contentObj, function(err, canManage) {
                 if (err) {
                     return callback(err);
                 }
 
-                // Apply the permission changes to the content. This will then update the content libraries as well
-                _setContentPermissions(ctx, contentObj, permissionChanges, function(err, newContentObj) {
-                    if (err) {
-                        return callback(err);
-                    }
-
-                    // If setting the content permissions results in any new members, we should insert them into the content members library
-                    ContentMembersLibrary.insert(contentObj, newMembers, function(err) {
-                        if (err) {
-                            log().warn({
-                                'err': err,
-                                'contentId': contentObj.id,
-                                'principalIds': newMemberIds
-                            }, 'An error occurred while inserting principals into content members library while setting content permissions');
+                AuthzAPI.getDirectRoles(_.keys(permissionChanges), contentObj.id, function(err, existingRoles) {
+                    if (!canManage) {
+                        if (_.contains(permissionChanges, false)) {
+                            return callback({'code': 401, 'msg': 'Only managers can remove access'});
                         }
 
-                        // If setting the content permissions results in removing members from the content item, we should remove them from the content members library
-                        ContentMembersLibrary.remove(contentObj, removedMemberIds, function(err) {
+                        var settingHigherRole = _.find(permissionChanges, function(role, principalId) {
+                            return ContentConstants.roles.ALL_PRIORITY.indexOf(effectiveRole) < ContentConstants.roles.ALL_PRIORITY.indexOf(role);
+                        });
+                        if (settingHigherRole) {
+                            return callback({'code': 401, 'msg': 'You cannot set another user\'s role higher than your own'});
+                        }
+                        var hasHigherRole = _.find(existingRoles, function(role, principalId) {
+                            return ContentConstants.roles.ALL_PRIORITY.indexOf(effectiveRole) < ContentConstants.roles.ALL_PRIORITY.indexOf(role);
+                        });
+                        if (hasHigherRole) {
+                            return callback({'code': 401, 'msg': 'You cannot change the role of someone with a higher role than your own'});
+                        }
+                    }
+
+                    // Ensure that we're not crossing tenant boundaries or violating principal visibility settings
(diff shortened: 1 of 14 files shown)


nicolaasmatthijs:

node_modules/oae-content/lib/api.js
```diff
@@ -1079,48 +1079,75 @@ var setContentPermissions = module.exports.setContentPermissions = function(ctx,
             return callback(err);
         }
 
-        // Check if the current user can manage the content
-        ContentAuthz.canManageContent(ctx, contentObj, function(err, canManage) {
+        // Check that the user isn't trying to assign any role higher than his own
```
mrvisser: `... than their own`

node_modules/oae-content/lib/api.js
```diff
@@ -1079,48 +1079,75 @@ var setContentPermissions = module.exports.setContentPermissions = function(ctx,
...
+                        if (settingHigherRole) {
+                            return callback({'code': 401, 'msg': 'You cannot set another user\'s role higher than your own'});
+                        }
+                        var hasHigherRole = _.find(existingRoles, function(role, principalId) {
+                            return ContentConstants.roles.ALL_PRIORITY.indexOf(effectiveRole) < ContentConstants.roles.ALL_PRIORITY.indexOf(role);
+                        });
+                        if (hasHigherRole) {
+                            return callback({'code': 401, 'msg': 'You cannot change the role of someone with a higher role than your own'});
+                        }
                     }
```
mrvisser: I think we can have a `ContentAuthz.canSetContentPermissions` that takes care of doing the `resolveEffectiveRole` check and the role promotion checks to ensure roles aren't being set higher than one's own role. A few notes about this process:

* The documentation of the method should be clear that it's only checking permissions surrounding access to the content item and the role promotions, and it's *not* checking permissions against the users being promoted (e.g., it won't catch if the user is trying to add a loggedin user from another tenant). That stuff is done in `checkNewContentPermissions`. Ideally it *would* do this, but it's probably too big of a bite to take right now.
* We don't need to do the `ContentAuthz.canManageContent` check. Instead, we can look at the effective role that came out of `resolveEffectiveRole` and see if `effectiveRole === ContentConstants.roles.MANAGER`

mrvisser: Additionally, can you rename `checkNewContentPermissions` to `checkTargetUserAccess` ? I think it's a bit more representative of what's going on in that method and doesn't imply that it will be looking at the role promotion issues.

node_modules/oae-content/tests/test-content.js
```diff
@@ -3160,17 +3161,34 @@ describe('Content', function() {
...
-                    ContentTestUtil.assertUpdateContentMembersFails(contexts['nicolaas'].restContext, contexts['stuart'].restContext, contentObj.id, members, 401, function() {
-                        ContentTestUtil.assertShareContentSucceeds(contexts['nicolaas'].restContext, contexts['stuart'].restContext, contentObj.id, _.keys(members), callback);
+                    members[contexts['ian'].user.id] = 'editor';
+                    // Editor can add new viewers and editors
+                    ContentTestUtil.assertUpdateContentMembersSucceeds(contexts['nicolaas'].restContext, contexts['stuart'].restContext, contentObj.id, members, function() {
+                        members = {};
+                        members[contexts['anthony'].user.id] = 'manager';
+                        // Editor cannot promote a user to manager
+                        ContentTestUtil.assertUpdateContentMembersFails(contexts['nicolaas'].restContext, contexts['stuart'].  restContext, contentObj.id, members, 401, function() {
+                            // Editor cannot remove anyone
```
mrvisser: If an editor can add a viewer, shouldn't they be able to remove a viewer? I think they should be able to undo a mistake they made if they added the wrong user, for example.

/cc @nicolaasmatthijs

mrvisser:
Left a few items of feedback, back to @stuartf

node_modules/oae-content/tests/test-content.js
stuartf: I thought this was strange too

node_modules/oae-content/lib/authz.js
```diff
@@ -119,3 +119,36 @@ var canViewAllContent = module.exports.canViewAllContent = function(ctx, content
         canViewAllContent(ctx, contentItems, callback, _illegalContentItems);
     });
 };
+
+var canSetContentPermissions = module.exports.canSetContentPermissions = function(ctx, contentObj, permissionChanges, callback) {
```
mrvisser: Needs docs

node_modules/oae-content/lib/authz.js
```diff
@@ -119,3 +119,36 @@ var canViewAllContent = module.exports.canViewAllContent = function(ctx, content
...
+                    return callback(null, false, 'You cannot set another user\'s role higher than your own');
+                }
+                var hasHigherRole = _.find(existingRoles, function(role, principalId) {
+                    return ContentConstants.roles.ALL_PRIORITY.indexOf(effectiveRole) < ContentConstants.roles.ALL_PRIORITY.indexOf(role);
+                });
+                if (hasHigherRole) {
+                    return callback(null, false, 'You cannot change the role of someone with a higher role than your own');
+                }
+            }
+            callback(null, true);
```
mrvisser: `return callback(null, true);`

mrvisser:
Added a couple small things, back to @stuartf for followup.

@stuartf I don't think this needs to be blocked by feedback about editors removing viewers. We can probably merge and iterate on it if necessary.

It looks like there is still follow-up needed in 3akai-ux, though.

stuartf:
Fixed review issues, still working on adding `role` parameter to the `shareContent` function

node_modules/oae-content/tests/test-content.js
nicolaasmatthijs: Sharing content behaves in exactly the same way. You're able to share an item with someone, but it's not possible to unshare the item.

mrvisser: @nicolaasmatthijs It just felt like giving them access to edit was a detail that would be a little bit more important to distinguish from standard content share functionality.

nicolaasmatthijs: That's fair point. Alternatively, we could copy Google Doc's behaviour as well, but that would basically mean that the only thing editors can't do is delete the item. Would it still be worth it at that point?

mrvisser: It depends on where the requirement is coming from. Are people asking for `editor` because they want people to edit but just not to delete the content item? Or are people worried that they'll "steal" the item by updating roles? I'm not exactly sure where the need for hierarchy is coming from.

nicolaasmatthijs: The requirement comes from some of the writing courses that we are supporting, where students are supposed to be able to contribute to writing a document without being able to manage it.

nicolaasmatthijs: @stuartf : I'm really sorry for this, but lets revert back to editors only being able to add viewers through sharing. Any chance we can get that done today so we can include this in the upcoming release?

nicolaasmatthijs: In other words, the only thing an editor should be able to do, is edit the document

mrvisser: Do they really need to add new editors, then? That's considered managing I would think. Maybe aside from editing, they are just limited to the regular viewer permissions such as share (only if the document isn't private), etc...

mrvisser: I wrote the previous comment before I saw the other comments, sorry.

stuartf:
This is ready for more review

nicolaasmatthijs:
@mrvisser @stuartf : Given the decision that was made yesterday, can this PR be closed?

nicolaasmatthijs:
Closing as this is no longer required

PR #1166, opened 2015-04-21
Enhance configurability of footer 

nicolaasmatthijs:
Example

```
*Unity by <*Research>, powered by the Apereo <Open Academic Environment software>. <Feedback> , <Terms & Conditions>
```

### DIFF
--- a/node_modules/oae-tenants/config/config.js
+++ b/node_modules/oae-tenants/config/config.js
@@ -67,8 +67,8 @@ module.exports = {
         'elements': {
             'instanceName': new Fields.Text('Instance name', 'The name of the current instance', '', {'tenantOverride': false}),
             'instanceURL': new Fields.Text('Instance URL', 'The URL of the main instance website', '', {'tenantOverride': false}),
-            'hostingOrganisation': new Fields.Text('Hosting organisation', 'The name of the organisation hosting the current instance', '', {'tenantOverride': false}),
-            'hostingOrganisationURL': new Fields.Text('Hosting organisation URL', 'The URL of the hosting organisation website', '', {'tenantOverride': false})
+            'hostingOrganization': new Fields.Text('Hosting organization', 'The name of the organization hosting the current instance', '', {'tenantOverride': false}),
+            'hostingOrganizationURL': new Fields.Text('Hosting organization URL', 'The URL of the hosting organization website', '', {'tenantOverride': false})
         }
     },
     'actions': {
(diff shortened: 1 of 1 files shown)


nicolaasmatthijs:

node_modules/oae-tenants/config/config.js
```diff
@@ -61,6 +61,16 @@ var timezoneConfigValues = _.chain(TZ.getZones())
...
+    'instance': {
+        'name': 'Instance Information',
+        'description': 'Information about the current instance',
+        'elements': {
+            'instanceName': new Fields.Text('Instance name', 'The name of the current instance', '', {'tenantOverride': false}),
+            'instanceURL': new Fields.Text('Instance URL', 'The URL of the main instance website', '', {'tenantOverride': false}),
+            'hostingOrganisation': new Fields.Text('Hosting organisation', 'The name of the organisation hosting the current instance', '', {'tenantOverride': false}),
+            'hostingOrganisationURL': new Fields.Text('Hosting organisation URL', 'The URL of the hosting organisation website', '', {'tenantOverride': false})
+        }
+    },
```
mrvisser: Hmm, being consistent on British v.s. American spellings might be something we want to do for non-i81n-able stuff, but I don't think there is much use fussing about it as part of this PR. There is already a mix in our config language.

mrvisser: Although, the config property names (Organisation) might be confusing. I feel like we've normally used American spelling for API things and variable/method names, but maybe it hasn't actually come up very often.

What do you think of UK-spelling config keys?

mrvisser: A quick check shows we don't have much of a precedence at least in config keys. I feel like using american spelling for dev-facing public API things makes usage more predictable, but it could just be because I've developed more in NA. What do you think?

mrvisser:
Back to @nicolaasmatthijs for a bit of feedback.

node_modules/oae-tenants/config/config.js
nicolaasmatthijs: I'm not too bothered either way, but I'll switch to American spelling for predictability.

nicolaasmatthijs:
Followed up. Back to @mrvisser for review.

davidoae:
Use US spelling as that's already what the rest of the world is getting
used to.
It'll only piss off the English. the ones that actually know the British
way, heh.

D

PR #1167, opened 2015-04-22
Add configuration options for horizontal and vertical alignment in landing page blocks

nicolaasmatthijs:
Add configuration options for horizontal and vertical alignment in landing page blocks


nicolaasmatthijs:

mrvisser:
Should we just allow a tenant admin to persist their own template(s) / CSS for landing page blocks?

nicolaasmatthijs:
@mrvisser : I don't think so. That's going to make it harder to wrap an easier admin UI around the configurable landing pages, which is something on our todo list. We might want to reconsider if we keep adding new styling config values, but I think it's fine for now.

Issue #1168, opened 2015-04-24
Email regressions


nicolaasmatthijs:
- The footer needs to be updated to reflect the new UI footer
- The styling is a bit messed up:

![screen shot 2015-04-24 at 09 01 42](https://cloud.githubusercontent.com/assets/109850/7323073/fda5af30-ea60-11e4-8d2b-f944243a58af.png)

nicolaasmatthijs:
Assigning to @mrvisser for fix

mrvisser:
@nicolaasmatthijs I submit a fix for the footer issues, but I can't reproduce the template getting messed up, though it happened to me on QA0 as well. I'd like to get the footer updates merged and then see if I can reproduce on QA0 again afterward.

mrvisser:
Provided fix for templating issue as part of https://github.com/oaeproject/Hilary/pull/1171

PR #1170, opened 2015-04-24
(Issue 1168) Update footer to match UI

mrvisser:
Fix for the footer issues in #1168 

Assigned to @nicolaasmatthijs for review

### DIFF
--- a/node_modules/oae-activity/emailTemplates/mail.html.jst
+++ b/node_modules/oae-activity/emailTemplates/mail.html.jst
@@ -36,11 +36,23 @@
         var oaeNameI18n = util.i18n.translate('__MSG__OPEN_ACADEMIC_ENVIRONMENT__');
         var url = baseUrl + '/me?emailpreferences';
 
-        var oaeLink = '<a href="http://www.oaeproject.org">' + oaeNameI18n + '</a>';
-        var instanceLink = (instance.URL) ? '<a href="' + instance.URL + '">' + util.html.encodeForHTML(instance.name) + '</a>' :
-            util.html.encodeForHTML(instance.name);
-        var hostingOrganizationLink = (hostingOrganization.URL) ? '<a href="' + hostingOrganization.URL + '">' + util.html.encodeForHTML(hostingOrganization.name) + '</a>' :
-            util.html.encodeForHTML(hostingOrganization.name);
+        var oaeLink = '<a href="http://www.oaeproject.org">' + util.html.encodeForHTML(oaeNameI18n) + '</a>';
+
+        var instanceLink = util.html.encodeForHTML(instance.name);
+        if (instance.URL) {
+            instanceLink =
+                '<a href="' + util.html.encodeForHTMLAttribute(instance.URL) + '">' +
+                    instanceLink +
+                '</a>';
+        }
+
+        var hostingOrganizationLink = util.html.encodeForHTML(hostingOrganization.name);
+        if (hostingOrganization.URL) {
+            hostingOrganizationLink =
+                '<a href="' + util.html.encodeForHTMLAttribute(hostingOrganization.URL) + '">' +
+                    hostingOrganizationLink +
+                '</a>';
+        }
 
         var footerContentData = {
             'oaeLink': oaeLink,
--- a/node_modules/oae-activity/emailTemplates/mail.html.jst
+++ b/node_modules/oae-activity/emailTemplates/mail.html.jst
@@ -33,15 +33,37 @@
      * Render the email footer
      */
     var renderFooter = function() {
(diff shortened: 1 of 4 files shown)


mrvisser:
Assigning back to self until travis runs, and discussion about remainder of the issue

mrvisser:
Assigning to @nicolaasmatthijs for review

node_modules/oae-activity/emailTemplates/mail.html.jst
```diff
@@ -33,15 +33,37 @@
      * Render the email footer
      */
     var renderFooter = function() {
-        var copyrightSymbolI18n = util.i18n.translate('__MSG__COPYRIGHT_SYMBOL__');
-        var apereoFoundationI18n = util.i18n.translate('__MSG__APEREO_FOUNDATION__');
+        var oaeNameI18n = util.i18n.translate('__MSG__OPEN_ACADEMIC_ENVIRONMENT__');
         var url = baseUrl + '/me?emailpreferences';
 
+        var oaeLink = '<a href="http://www.oaeproject.org">' + oaeNameI18n + '</a>';
+        var instanceLink = (instance.URL) ? '<a href="' + instance.URL + '">' + util.html.encodeForHTML(instance.name) + '</a>' :
```
nicolaasmatthijs: Can we do the same as the footer widget? I'm not a huge fan of multi-line tertiary statements

nicolaasmatthijs:
```
var instanceLink = instance.name;
if (instance.URL) {
  instanceLink = '<a href="' + instance.URL + '">' + instance.name + '</a>';
}

File: node_modules/oae-activity/tests/test-email.js [hunk 2]
```diff
@@ -135,6 +135,112 @@ describe('Activity Email', function() {
...
+                    ConfigTestUtil.clearConfigAndWait(globalAdminRestContext, null, _.union(instanceConfigFields, hostingOrganizationConfigFields), function(err) {
+                        assert.ok(!err);
+
+                        // Create a link, let Nico manage it
+                        RestAPI.Content.createLink(mrvisser.restContext, 'Google', 'Google', 'public', 'http://www.google.ca', [nico.user.id], [], [], function(err, link) {
+                            assert.ok(!err);
+
+                            // Ensure the email is as expected
+                            EmailTestsUtil.collectAndFetchAllEmails(function(messages) {
+                                assert.notEqual(messages[0].html.indexOf('Apereo <a href=\"http://www.oaeproject.org\" '), -1);
```
nicolaasmatthijs: Rather than repeating this everywhere, can we do an `assertEmailFooter` function that checks all of these?

nicolaasmatthijs:
Finished review. Back to @mrvisser for follow-up

node_modules/oae-activity/tests/test-email.js
mrvisser: I'm not sure such a function would be very re-usable outside this test. Can you give an example of what the function signature would be and where it would sit?

nicolaasmatthijs: It could just sit above this test. I guess it would be something like this:

```
assertEmailFooter(message[0].html, {'expectInstanceName': true, 'expectInstanceURL': true, 'expectHostingOrganization': true, 'expectHostingOrganizationURL': true});
```

node_modules/oae-activity/emailTemplates/mail.html.jst
mrvisser: The suggested code can't be done because the `instance.name` has to be run through an `encodeForHTML` function. I feel using a tertiary statement is the most easy to read version of this logic.

nicolaasmatthijs: Well, it would just be

```
var instanceLink = util.html.encodeForHTML(instance.name);
if (instance.URL) {
  instanceLink = '<a href="' + instance.URL + '">' + util.html.encodeForHTML(instance.name) + '</a>';
}
```

nicolaasmatthijs: Note that this mimics the UI code as well

mrvisser: Didn't realize `encodeForHtml` wouldn't fail if it received `null`/`undefined`.

Note that it doesn't mimic the UI as the UI doesn't seem to encode for HTML, which is probably a bug. I'll try and verify.

nicolaasmatthijs:
Left 1 more comment. Back to @mrvisser

mrvisser:
Added another test, back to @nicolaasmatthijs

nicolaasmatthijs:
Left some final comments. Back to @mrvisser .

mrvisser:
Fixed

nicolaasmatthijs:
Looks good. Merged.

PR #1173, opened 2015-04-27
Use : instead of - in strategy id

mrvisser:


node_modules/oae-tenants/tests/test-tenants.js
```diff
@@ -445,6 +445,28 @@ describe('Tenants', function() {
...
+         */
+        it('verify creating a tenant with a dash in the alias does not break authentication strategy', function(callback) {
+            var tenantHost = TestsUtil.generateRandomText(1);
+            var tenantAlias = 'test-with-dash';
+            TestsUtil.createTenantWithAdmin('alias-with-dash', tenantHost, function(err, tenant, tenantAdminRestContext) {
+                assert.ok(!err);
+
+                // Ensure the tenant admin's me object propertly represents the authentication
+                // strategy
+                RestAPI.User.getMe(tenantAdminRestContext, function(err, me) {
```
nicolaasmatthijs: @mrvisser : I know this is almost definitely the same, but would you mind verifying this with a regular user instead of an admin user?

nicolaasmatthijs:
Left 1 comment, should be good to be merged after that. Back to @mrvisser for follow-up.

mrvisser:
Fixed, back to @nicolaasmatthijs

nicolaasmatthijs:
Looks good. Merged.

PR #1178, opened 2015-04-28
(Issue 1175) New-line before every HTML attribute

mrvisser:
I'm currently testing this, but thought I'd run the approach by you.

It adds a new line before all HTML attributes. It's quite aggressive, but should allow us to mostly forget about line length, unless our inlined styles get crazy.

As things stand right now with this fix, the longest email line that gets generated in our unit tests is ~350 chars long.

### DIFF
--- a/node_modules/oae-activity/tests/test-email.js
+++ b/node_modules/oae-activity/tests/test-email.js
@@ -21,7 +21,8 @@ var util = require('util');
 
 var AuthzUtil = require('oae-authz/lib/util');
 var ConfigTestUtil = require('oae-config/lib/test/util');
-var EmailTestsUtil = require('oae-email/lib/test/util');
+var EmailTestUtil = require('oae-email/lib/test/util');
+var PrincipalsTestUtil = require('oae-principals/lib/test/util');
 var RestAPI = require('oae-rest');
 var RestContext = require('oae-rest/lib/model').RestContext;
 var Sanitization = require('oae-util/lib/sanitization');
@@ -55,7 +56,7 @@ describe('Activity Email', function() {
         globalAdminRestContext = TestsUtil.createGlobalAdminRestContext();
 
         // Flush the pending mails
-        EmailTestsUtil.clearEmailCollections(function() {
+        EmailTestUtil.clearEmailCollections(function() {
             refreshConfiguration(null, false, false, {}, function(config) {
                 return callback();
             });
@@ -146,7 +147,7 @@ describe('Activity Email', function() {
                 assert.ok(!err);
 
                 // Clear emails to start
-                EmailTestsUtil.collectAndFetchAllEmails(function() {
+                EmailTestUtil.collectAndFetchAllEmails(function() {
                     var instanceName = TestsUtil.generateRandomText(1);
                     var instanceURL = util.format('http://www.instance.oaeproject.org/%s', TestsUtil.generateRandomText(1));
                     var hostingOrganization = TestsUtil.generateRandomText(1);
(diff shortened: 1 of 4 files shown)


node_modules/oae-activity/tests/test-email.js
```diff
@@ -1657,4 +1658,42 @@ describe('Activity Email', function() {
...
+                // Add the members to the group
+                var members = {};
+                members[nico.user.id] = 'manager';
+                PrincipalsTestUtil.assertSetGroupMembersSucceeds(simong.restContext, group.id, members, function() {
+
+                    // Collect any email notifications so we can focus on just the one we generate next
+                    EmailTestUtil.collectAndFetchAllEmails(function() {
+
+                        // Create 2 links for the group
+                        RestAPI.Content.createLink(simong.restContext, 'Google', 'Google', 'public', 'http://www.myawesomelinkthatilike.ca', [group.id], [], [], function(err) {
```
simong: I know it's unrealistic, but should we have a test where a piece of content has a displayName with more than a 1000 characters and ensure everything is properly trimmed?

mrvisser: This PR won't fix that as it is right now. We can't take a "container" approach to that, it would have to be handled in the template generation

node_modules/oae-email/lib/test/util.js
```diff
@@ -106,13 +106,17 @@ var collectAndFetchAllEmails = module.exports.collectAndFetchAllEmails = functio
...
+                        // ensures no email has an html content line longer than 500 chars, which
+                        // accounts for post-processing things such as SendGrid changing the links
+                        // to extremely long values (e.g., ~400 characters long) for click-tracking
+                        //
                         // @see https://github.com/oaeproject/Hilary/issues/1168
                         _.each(messages, function(message) {
                             assert.ok(_.isString(message.html));
                             _.each(message.html.split('\n'), function(line) {
-                                assert.ok(line.length <= 997, util.format('Expected no email line to be more than 997 characters, but found: %s', line));
+                                assert.ok(line.length <= 500, util.format('Expected no email line to be more than 500 characters, but found: (%s) %s', line.length, line));
```
simong: Should we check that a line can't have more than 2 links as sendgrid might push it over 1k?

simong:
Left a couple of questions for @mrvisser

mrvisser:
I've added a change that ensures there is only 1 link per line. I've also tested some emails on QA0, but we probably want to bash this before released.

Note that we would either have to three-dot the display names at some length (50?) to ensure that they don't push above our line limit. Since we use display names in `title` attributes, wrapping to a new line is not an option.

node_modules/oae-email/lib/api.js
```diff
@@ -399,7 +399,19 @@ var sendEmail = module.exports.sendEmail = function(templateModule, templateId,
             return callback(err);
         }
 
-        emailInfo.html = inlinedHtml;
+        // Process the HTML such that we add line breaks before each html attribute to try and keep
```
simong: **after** each tag name and attribute ?

simong:
PR LGTM, As you said, we probably want to bugbash this first

node_modules/oae-email/lib/api.js
mrvisser: No, only before each attribute, strictly speaking

mrvisser:
@simong I don't think we need to bugbash before merging, I think we should bugbash before release. We can merge this, then bugbash it on QA0 in our regular release workflow.

mrvisser:
Put back into `To Review` and back to @simong

PR #1179, opened 2015-04-29
Massage config for tests so they can run on a release server

mrvisser:
Enforce some of the test configuration assumptions we make.


mrvisser:
:blush:

PR #1152, opened 2015-05-06
Deprecated usage of strftime

simong:
strftime has deprecated one of their methods. When running the unit tests I get:

> [WARNING] `require('strftime').strftimeTZ(format, date, tz)` is deprecated and will be removed in version 1.0. Instead, use `var s = require('strftime').timezone(tz); s(format, [date])` or `require('strftime').timezone(tz)(format, [date])`.


nicolaasmatthijs:
Looks good. Merged

PR #1182, opened 2015-05-06
(TravisCI) Output VM memory stats after the tests

mrvisser:


mrvisser:
Looks to be giving the output we need. Assigning to @simong for review.

mrvisser:
Note that the intent of this is to try and do post-mortem analysis on a regular test failure where the logs indicate that cassandra and redis both stopped responding. This could be a result of a resource issue on the VM.

Issue #1175, opened 2015-04-28
Certain activities still generate emails with lines with more than a 1000 characters


simong:
The `ACTIVITY_CONTENT_CREATE_2+_GROUP` activity for example generates the following summary: `<user> created <content1> and <content2> in the group <group>.`

Because of the CSS inlining each `<a>` tag gets about 150 characters added on top of it. This quickly blows up and crosses the postfix 1000characters threshold (even before sendgrid injects its links).

Additionally, there's a `<div class="wrapped"` on the same line that pushes that line count a bit further.

nicolaasmatthijs:
Assigning to @mrvisser for further investigation as this is currently a production problem.

mrvisser:
PR at #1178

PR #1183, opened 2015-05-06
Use the latest version of juice

stuartf:
this fix incorporates #1175


stuartf:
This is ready for review

mrvisser:
Looks good. We need to bug bash emails with this, though.

PR #1187, opened 2015-05-12
(UserDelete) Implement the ability to delete users

mrvisser:


mrvisser:

nicolaasmatthijs:
@mrvisser : Would you also be able to create a corresponding UI PR that adds the user delete functionality to the administration UI? The idea is that a third action would be added to the "Action" tab, showing a red "Delete x" button that would trigger a confirmation dialog before deleting the user.

screen shot 2015-05-12 at 23 54 10

mrvisser:
@nicolaasmatthijs will do! I just added the admin server route for this to support it, so that's why a new commit came in @simong . Still ready for review.

mrvisser:
Note this depends on https://github.com/oaeproject/oae-rest/pull/16

node_modules/oae-authz/lib/util.js
```diff
@@ -16,6 +16,7 @@
 var _ = require('underscore');
 
 var AuthzConstants = require('oae-authz/lib/constants').AuthzConstants;
+var AuthzDelete = require('oae-authz/lib/delete');
```
simong: This file doesn't seem to be committed?

simong: Nvm, I'm clearly an idiot

node_modules/oae-principals/lib/internal/dao.js
```diff
@@ -402,6 +361,87 @@ var isUser = module.exports.isUser = function(userId) {
...
+var _deletePrincipalFields = function(principalId, profileFields, callback) {
+    // Remove the specified fields
+    var query = util.format('DELETE "%s" FROM "Principals" where "principalId" = ?', profileFields.join('", "'));
+    Cassandra.runQuery(query, [principalId], function(err) {
+        if (err) {
+            return callback(err);
+        }
+
+        // If the principal is a user, invalidate their cache entry
+        OaeUtil.invokeIfNecessary(isUser(principalId), invalidateCachedUsers, [principalId], function(err) {
```
simong: At first I thought this line could be removed because we're calling `_updatePrincipal` which does some cache handling.
However, that isn't true as we're removing principal fields here and `_updatePrincipal` will only update specific profile fields, not refresh the entire object.

So, could you add a bit of documentation why `invalidateCachedUsers` is required here?

simong:
Can you track deleting a user with mixpanel?

simong:
LGTM, just 2 minor things. Back to @mrvisser

mrvisser:
Followed up, back to @simong . I don't think this needs to wait for the UI portion

mrvisser:
Fixed typo

PR #1191, opened 2015-05-14
Added a script that deletes the session keys from the Etherpad Column family

simong:
We're currently creating a new session everytime you join a pad. This didn't used to be a problem in 1.4, but it seems that 1.5.6 pulls down all the sessions for a user and stores them again. This takes a bit too long sometimes.

This PR removes all the sessions in etherpad. This is only a stop-gap measure, as we should have something that binds only 1 session to a user/pad.

### DIFF
--- a/etc/tasks/deleteEtherpadSessions.js
+++ b/etc/tasks/deleteEtherpadSessions.js
@@ -1,0 +1,78 @@
+/*!
+ * Copyright 2015 Apereo Foundation (AF) Licensed under the
+ * Educational Community License, Version 2.0 (the "License"); you may
+ * not use this file except in compliance with the License. You may
+ * obtain a copy of the License at
+ *
+ *     http://opensource.org/licenses/ECL-2.0
+ *
+ * Unless required by applicable law or agreed to in writing,
+ * software distributed under the License is distributed on an "AS IS"
+ * BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express
+ * or implied. See the License for the specific language governing
+ * permissions and limitations under the License.
+ */
+
+var _ = require('underscore');
+var util = require('util');
+
+var Cassandra = require('oae-util/lib/cassandra');
+var log = require('oae-logger').logger('delete-etherpad-sessions');
+var OAE = require('oae-util/lib/oae');
+
+// The application configuration
+var config = require('../../config').config;
+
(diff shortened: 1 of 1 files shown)


etc/tasks/deleteEtherpadSessions.js
```diff
@@ -0,0 +1,78 @@
+/*!
```
nicolaasmatthijs: Is this script something that we'll need to run manually from time to time? How difficult would it be to make sure that the sessions are cleaned up properly?

etc/tasks/deleteEtherpadSessions.js
```diff
@@ -0,0 +1,78 @@
...
+
+var _ = require('underscore');
+var util = require('util');
+
+var Cassandra = require('oae-util/lib/cassandra');
+var log = require('oae-logger').logger('delete-etherpad-sessions');
+var OAE = require('oae-util/lib/oae');
+
+// The application configuration
+var config = require('../../config').config;
```
nicolaasmatthijs: Should we keep into account that a custom config file can be specified in app.js ?

etc/tasks/deleteEtherpadSessions.js
```diff
@@ -0,0 +1,78 @@
...
+ * @api private
+ */
+var _deleteSessionRows = function(rows, callback) {
+    // Get the session keys
+    var keysToDelete = _.chain(rows)
+        .map(function(row) {
+            return row.get('key').value;
+        })
+        .filter(function(key) {
+            return (key.indexOf('session') !== -1);
```
nicolaasmatthijs: What is this trying to catch?

nicolaasmatthijs:
Finished review. Back to @simong for follow-up

etc/tasks/deleteEtherpadSessions.js
simong: We only want keys with the string `session` in them

etc/tasks/deleteEtherpadSessions.js
simong: For now it is, we should persist the sessionID for a user/pad and re-use it on subsequent joins. That's going to require some more work and thinking though.

simong:
Followed up on review, back to @nicolaasmatthijs

nicolaasmatthijs:
Looks good. Merged.

PR #1188, opened 2015-05-27
Remove backward compatibility route for members library search: `members`

mrvisser:
STEM incubator UI is now transitioned over as per @sathomas


mrvisser:
Assigned to @simong for review.

PR #1192, opened 2015-06-02
Documentation warning

stuartf:
Fresh install of OAE on OS X, first run, has the following in the log:

```
[2015-05-27T16:16:05.319Z]  WARN: oae-doc/15618 on lawn-143-215-206-78.lawn.gatech.edu: Failed parsing comment data with dox for file /Users/stephen/Documents/Development/oae/Hilary/node_modules/oae-util/lib/../../../node_modules/oae-tenants/lib/api.js. Ignoring.
    err: {
      "name": "TypeLexerSyntaxError",
      "message": "Expected \"!\", \"(\", \"*\", \"...\", \"=\", \"?\", \"function\", \"module\", \"{\" or [a-zA-Z_$] but \"[\" found.:\n[Function]\n^"
    }
```

Presumably that's from

```
    /**
     * Disable or enable a tenant
     *
     * @param  {Context}      ctx             Standard context object containing the current user and the current tenant
     * @param  {String[]}     aliases         An array of aliases representing the tenants that should be stopped
     * @param  {Boolean}      disabled        True if the tenant needs to be disabled
     * @param  {[Function]}   callback        Callback function executed when request is completed
     * @param  {Object}       callback.err    An error that occurred, if any
     */
```

### DIFF
--- a/node_modules/oae-principals/lib/delete.js
+++ b/node_modules/oae-principals/lib/delete.js
@@ -168,7 +168,7 @@ var _invokeGroupHandlers = function(log, handlers, group) {
  * @param  {Logger}         log         The logger to use to report progress
  * @param  {Object}         handlers    The handler functions keyed by their handler name, indicating which handlers to invoke
  * @param  {User|Group}     principal   The user or group that was the target of the operation
- * @param  {Object...}      args...     A variable number of arguments for the handler depending on its type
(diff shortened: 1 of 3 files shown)


stuartf:
fixed review issue

mrvisser:
@stuartf I don't think you pushed the review follow-up

mrvisser:
Looks good

PR #1193, opened 2015-06-03
Update strategies.js for Preventing modification of the name and mail for CAS and Shibb

dooremont:
Preventing modification of the name and mail for CAS and Shibb

### DIFF
--- a/node_modules/oae-authentication/config/strategies.js
+++ b/node_modules/oae-authentication/config/strategies.js
@@ -58,6 +58,7 @@ module.exports = {
         'description': 'Allow Shibboleth authentication for tenant',
         'elements': {
             'enabled': new Fields.Bool('Shibboleth Authentication Enabled', 'Allow Shibboleth authentication for tenant', false),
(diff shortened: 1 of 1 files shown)


node_modules/oae-authentication/config/strategies.js
```diff
@@ -58,6 +58,7 @@ module.exports = {
         'description': 'Allow Shibboleth authentication for tenant',
         'elements': {
             'enabled': new Fields.Bool('Shibboleth Authentication Enabled', 'Allow Shibboleth authentication for tenant', false),
+            'disableProfile': new Fields.Bool('Preventing modification of the name and mail', 'Preventing modification of the name and mail', true),
```
nicolaasmatthijs: @dooremont : Is it necessary to disable modification of the email as well? What if CAS has returned an invalid email address?

nicolaasmatthijs:
Left a question for @dooremont

dooremont:
Normally in our operation that we check to IT then sends out the information into the administrators can prevent changes

dooremont:
do you think added this modification for the next release ? 
Left a question for @nicolaasmatthijs

dooremont:
replace by  #1313

PR #1186, opened 2015-06-08
REST documentation out of date

mrvisser:
It looks like the Swagger-based REST documentation is out of date:

screen shot 2015-05-12 at 10 40 34 am

I'm not sure what `memberships-library` does, but I think `members` should be changed to `members-library`? 

https://github.com/oaeproject/Hilary/blob/master/node_modules/oae-search/lib/rest.js#L56


mrvisser:
Assigning to @sathomas for review

sathomas:
@mrvisser the fix looks fine to merge, but I'm not sure that the Travis CI should be ignored.

mrvisser:
@sathomas yes that one I believe is a known intermittent failure and unrelated to this fix. Those tests are being investigated as they occur on master builds, where we have access to the log files.

sathomas:
Merged

PR #1197, opened 2015-06-08
(TravisCI Experimenting) Tuning Cassandra JVM args and timeouts

mrvisser:


etc/cassandra/cassandra-env.sh
```diff
@@ -0,0 +1,271 @@
...
+        ;;
+    esac
+
+    # some systems like the raspberry pi don't report cores, use at least 1
+    if [ "$system_cpu_cores" -lt "1" ]
+    then
+        system_cpu_cores="1"
+    fi
+
+    echo "LOL setting max heap size any way"
```
simong: happy debugging?

mrvisser: Yea it was a hoot :) Last night I got the first passing build I've seen in a while. So, progress!

PR #1200, opened 2015-06-12
Replace 'interact' search with 'share' search

mrvisser:


mrvisser:
Assigning to @simong for review. This is a piece of inviting guests that we could merge to master (there will be a 3akai-ux PR to go along with this)

mrvisser:
Placing on hold as the complexities of sharing multiple resources at once are currently undecided

nicolaasmatthijs:
> @nicolaas: in the case where the user has selected a number of items from their library and chose to "Share", do we want it to intelligently determine which users the current user can actually share with?

I think it would make more sense if the user is able to see all users/groups they're able to share with, followed by an error message explaining that they're not able to share certain items with certain people/groups. If the returned user/group list is dependent on the items that are selected, users might be confused by the inconsistency and might think that the system is not working as expected.

mrvisser:
Closing based on feedback. That is how things currently work, except for better UI messaging which is a 3akai-ux concern.

PR #1201, opened 2015-06-13
Add REST endpoint to get login ids for a user

Coenego:
In order to support showing login ids for a user in the admin UI, an endpoint must be created to, given a user id, get the list of login ids associated to that user. This would require the following work:

1. There is an [internal `_getUserLoginIds` function](https://github.com/oaeproject/Hilary/blob/master/node_modules/oae-authentication/lib/api.js#L753-L783) that lists them from Cassandra. A new public function should be created that takes in the user id whose login ids to list and delegates to that internal function
    * Validation: Only admins of the specified user id or the user themself can get their list of login ids
2. A new GET request can be mounted in `oae-authentication/lib/rest.js` that delegates to this new public function. Maybe something like `GET /api/auth/loginIds/:userId` would work well, and in the future if we implement associating new login ids to user accounts, we could `POST /api/auth/loginIds/:userId/:strategyName`
3. Caution should be taken as the "local" auth strategy will have the password hash in its response. It should be omitted

### DIFF
--- a/node_modules/oae-authentication/lib/api.js
+++ b/node_modules/oae-authentication/lib/api.js
@@ -770,16 +770,24 @@ var getUserLoginIds = module.exports.getUserLoginIds = function(ctx, userId, cal
     PrincipalsAPI.getUser(ctx, userId, function(err, user) {
         if (err) {
             return callback(err);
-        }
-
-        // Only global administrators and administrators of the tenant the user belongs to can request the login ids
-        var isGlobalAdmin = (ctx.user() && ctx.user().isGlobalAdmin());
-        var isAdmin = (ctx.user().isAdmin(user.tenant.alias));
-        if (!isGlobalAdmin && !isAdmin) {
+        } else if (!ctx.user().isAdmin(user.tenant.alias)) {
+            // Only global administrators and administrators of the tenant the user belongs to can request the login ids
             return callback({'code': 401, 'msg': 'You are not authorized to request the login ids for this user'});
         }
 
-        return _getUserLoginIds(userId, callback);
+        return _getUserLoginIds(userId, function(err, _loginIds) {
+            if (err) {
+                return callback(err);
+            }
+
+            // Only return the strategies and their corresponding external id
+            var loginIds = {};
+            _.each(_loginIds, function(values, strategy) {
+                loginIds[strategy] = values.externalId;
+            });
+
+            return callback(null, loginIds);
+        });
     });
 };
 
--- a/node_modules/oae-authentication/lib/api.js
+++ b/node_modules/oae-authentication/lib/api.js
@@ -594,8 +594,6 @@ var changePassword = module.exports.changePassword = function(ctx, userId, oldPa
                 return callback({'code': 400, 'msg': 'User does not have a local account mapping'});
             }
 
-
-
             // Determine the current user access
             var isAdmin = ctx.user().isAdmin(localLoginId.tenantAlias);
             var isTargetUser = ctx.user().id === userId;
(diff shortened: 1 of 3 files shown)


Coenego:

node_modules/oae-authentication/lib/api.js
```diff
@@ -753,6 +751,41 @@ var _associateLoginId = function(loginId, userId, callback) {
...
+        if (err) {
+            return callback(err);
+        }
+
+        // Only global administrators and administrators of the tenant the user belongs to can request the login ids
+        var isGlobalAdmin = (ctx.user() && ctx.user().isGlobalAdmin());
+        var isAdmin = (ctx.user().isAdmin(user.tenant.alias));
+        if (!isGlobalAdmin && !isAdmin) {
+            return callback({'code': 401, 'msg': 'You are not authorized to request the login ids for this user'});
+        }
```
mrvisser: A few things about this check:

1. We don't have to check if `ctx.user()` is defined, because `isLoggedInUser` will have already ensured that
2. `ctx.user().isAdmin(user.tenant.alias)` will also take into consideration `isGlobalAdmin`, so we just need the one check
3. The check should be done in an `else if` of the prior error handling if statement

node_modules/oae-authentication/lib/api.js
```diff
@@ -753,6 +751,41 @@ var _associateLoginId = function(loginId, userId, callback) {
...
+
+        return _getUserLoginIds(userId, callback);
```
mrvisser: Since the user login ids will contain sensitive information (e.g., the password hash), we'll want to make sure we don't send it in the response.

We can probably just make the response keyed by strategy, and the value is the external id, and only return those values.

node_modules/oae-authentication/tests/test-auth.js
```diff
@@ -0,0 +1,107 @@
...
+                                    RestAPI.Authentication.getUserLoginIds(camAdminRestContext, createdUser.id, function(err, loginIds) {
+                                        assert.ok(!err);
+                                        assert.ok(loginIds);
+
+                                        // Verify that a global admin can request the login ids for the test user
+                                        RestAPI.Authentication.getUserLoginIds(globalAdminRestContext, createdUser.id, function(err, loginIds) {
+                                            assert.ok(!err);
+                                            assert.ok(loginIds);
+
+                                            callback();
```
mrvisser: The unit tests should include tests that verify the login ids we expect are returned in the response. To avoid having a single test that gets way too big, it would probably be worth having one `it` block for validation+authorization tests, and then one block for ensuring the validity of the login ids.

mrvisser: Also super minor, that `callback();` should be returned: `return callback();`

mrvisser:
Overall looks really good, just a security probe and some nagging on the tests. Back to @Coenego

Coenego:
Back to @mrvisser for review after follow-up

node_modules/oae-authentication/tests/test-auth.js
```diff
@@ -0,0 +1,130 @@
...
+
+        /**
+         * Test that verifies that a user's login ids can be requested
+         */
+        it('verify that only authorized admins can request a user\'s login ids', function(callback) {
+            // Generate a user id
+            var userId = TestsUtil.generateTestUserId();
+
+            // Verify that an error is thrown when a malformed user id was specified
+            RestAPI.Authentication.getUserLoginIds(globalAdminRestContext, userId, function(err, loginIds) {
```
mrvisser: We should make the user of `userId` here a more obvious malformed user id, such as "malformed user id", I'll follow up on this.

node_modules/oae-authentication/lib/api.js
```diff
@@ -753,6 +751,49 @@ var _associateLoginId = function(loginId, userId, callback) {
...
+    // Request the user details
+    PrincipalsAPI.getUser(ctx, userId, function(err, user) {
+        if (err) {
+            return callback(err);
+        } else if (!ctx.user().isAdmin(user.tenant.alias)) {
+            // Only global administrators and administrators of the tenant the user belongs to can request the login ids
+            return callback({'code': 401, 'msg': 'You are not authorized to request the login ids for this user'});
+        }
+
+        return _getUserLoginIds(userId, function(err, _loginIds) {
```
mrvisser: This shouldn't be `return`ed, I'll follow up

node_modules/oae-authentication/tests/test-auth.js
```diff
@@ -0,0 +1,130 @@
...
+                assert.ok(err);
+                assert.equal(err.code, 400);
```
mrvisser: Just FYI on a couple things I'm changing here as well that I noticed:

* We use `assert.strictEqual` instead of `assert.equal` now, and we change instances of `assert.equal` as we come across them. There wasn't really any way for you to know that though :P
* I've added a test both for invalid user ids: `'not an id'` and `'g:not:a-user-id'` to test that we both test that it is a valid id and specifically a user id

PR #1202, opened 2015-06-15
(InvitingGuests/TenantDomains) Add ability specify an email domain expression for a tenant

mrvisser:
Depends on oae-rest: oaeproject/oae-rest#20

### DIFF
--- a/node_modules/oae-tenants/config/config.js
+++ b/node_modules/oae-tenants/config/config.js
@@ -91,12 +91,5 @@ module.exports = {
         'elements': {
             'timezone': new Fields.List('Tenant Timezone', 'Tenant timezone', 'Etc/UTC', timezoneConfigValues, {'suppress': true})
         }
-    },
-    'domains': {
-        'name': 'Tenant domains',
-        'description': 'Domains associated to this tenant',
-        'elements': {
-            'email': new Fields.Text('Email domain', 'A regular expression that identifies the email domains associated to this tenant', '', {'tenantOverride': false, 'suppress': true, 'globalAdminOnly': true})
-        }
     }
 };
--- a/node_modules/oae-tenants/lib/api.js
+++ b/node_modules/oae-tenants/lib/api.js
@@ -263,7 +263,7 @@ var createTenant = module.exports.createTenant = function(ctx, alias, displayNam
     host = host.toLowerCase();
 
     // Ensure there are no conflicts
-    validator.check(null, {'code': 400, 'msg': 'This hostname is reserved'}).not(serverConfig.shibbolethSPHost);
+    validator.check(host, {'code': 400, 'msg': 'This hostname is reserved'}).not(serverConfig.shibbolethSPHost);
     validator.check(getTenant(alias), {'code': 400, 'msg': util.format('A tenant with the alias "%s" already exists', alias)}).isNull();
     validator.check(getTenantByHost(host), {'code': 400, 'msg': util.format('A tenant with the host "%s" already exists', host)}).isNull();
 
--- a/node_modules/oae-tenants/lib/api.js
+++ b/node_modules/oae-tenants/lib/api.js
@@ -171,14 +171,11 @@ var getTenantByHost = module.exports.getTenantByHost = function(tenantHost) {
  * @return {Tenant}                     The tenant that matches the domain, if any
  */
 var getTenantByEmailDomain = module.exports.getTenantByEmailDomain = function(emailDomain) {
-    var tenant = _.chain(getTenants())
-        .filter(function(tenant) {
-            // Filter the list of all tenants down to just the one (should be only one due to
-            // validation) whose email domain expression matches the specified email domain
-            return (tenant.emailDomain && _matchesEmailDomain(emailDomain.toLowerCase(), tenant.emailDomain));
-        })
-        .first()
-        .value();
+    var tenant = _.find(getTenants(), function(tenant) {
+        // Filter the list of all tenants down to just the one (should be only one due to
+        // validation) whose email domain expression matches the specified email domain
+        return (tenant.emailDomain && _matchesEmailDomain(emailDomain.toLowerCase(), tenant.emailDomain));
+    });
     return _copyTenant(tenant);
 };
 
@@ -270,9 +267,13 @@ var createTenant = module.exports.createTenant = function(ctx, alias, displayNam
     // Ensure only valid optional fields are set
     _.each(opts, function(val, key) {
         validator.check(key, {'code': 400, 'msg': util.format('Invalid field: %s', key)}).isIn(['emailDomain']);
-        if (key === 'emailDomain' && _.isString(val)) {
-            opts[key] = opts[key].toLowerCase();
-            _validateEmailDomain(validator, opts[key]);
+        if (key === 'emailDomain') {
+            validator.check(null, {'code': 400, 'msg': 'An email domain was passed in, but was not a string'}).isString(val);
+            if (!validator.hasErrors()) {
+                // Only continue if we know the email domain value was a string
+                opts[key] = opts[key].toLowerCase().trim();
+                _validateEmailDomain(validator, opts[key]);
+            }
         }
     });
     if (validator.hasErrors()) {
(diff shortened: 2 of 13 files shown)


mrvisser:

mrvisser:
Depends on oae-rest: https://github.com/oaeproject/oae-rest/pull/20

nicolaasmatthijs:
@mrvisser : Can you briefly explain how this PR changes things?

node_modules/oae-tenants/config/config.js
```diff
@@ -91,12 +91,5 @@ module.exports = {
         'elements': {
             'timezone': new Fields.List('Tenant Timezone', 'Tenant timezone', 'Etc/UTC', timezoneConfigValues, {'suppress': true})
         }
-    },
```
simong: Will this cause an issue with the data that's already in production? (probably not)

mrvisser: Only if it's used, which you told me it isn't :)

node_modules/oae-tenants/lib/api.js
```diff
@@ -166,6 +165,24 @@ var getTenantByHost = module.exports.getTenantByHost = function(tenantHost) {
 };
 
 /**
+ * Get a tenant whose configured email domain matches the specified email domain
+ *
+ * @param  {String}     emailDomain     The email domain with which to lookup the tenant
+ * @return {Tenant}                     The tenant that matches the domain, if any
+ */
+var getTenantByEmailDomain = module.exports.getTenantByEmailDomain = function(emailDomain) {
+    var tenant = _.chain(getTenants())
```
simong: This entire chain can be replaced by:

```javascript
var tenant = _.find(getTenants(), function(tenant) { 
    // Filter the list of all tenants down to just the one (should be only one due to
    // validation) whose email domain expression matches the specified email domain
    return (tenant.emailDomain && _matchesEmailDomain(emailDomain.toLowerCase(), tenant.emailDomain));
});
````

File: node_modules/oae-tenants/lib/api.js [hunk 3]
```diff
@@ -217,75 +238,78 @@ var _cacheTenants = function(callback) {
...
-            callback({'code': 400, 'msg': 'A tenant with the host ' + host + ' already exists'});
+    // Ensure there are no conflicts
+    validator.check(host, {'code': 400, 'msg': 'This hostname is reserved'}).not(serverConfig.shibbolethSPHost);
+    validator.check(getTenant(alias), {'code': 400, 'msg': util.format('A tenant with the alias "%s" already exists', alias)}).isNull();
+    validator.check(getTenantByHost(host), {'code': 400, 'msg': util.format('A tenant with the host "%s" already exists', host)}).isNull();
+
+    // Ensure only valid optional fields are set
+    _.each(opts, function(val, key) {
+        validator.check(key, {'code': 400, 'msg': util.format('Invalid field: %s', key)}).isIn(['emailDomain']);
+        if (key === 'emailDomain' && _.isString(val)) {
```
simong: If `val` is not a string, this validator won't run which might cause issues

mrvisser:
Just noticed this needs a migration CQL script to add `emailDomain` to the tenant table.

node_modules/oae-tenants/lib/api.js
```diff
@@ -494,6 +534,64 @@ var _setLandingPageBlockAttribute = function(ctx, block, blockName, attributeNam
...
+        'code': 400,
+        'msg': util.format('The email domain "%s" conflicts with existing email domain: %s', emailDomain, matchingEmailDomain)
+    }).isNull();
+};
+
+/**
+ * Determine if the given email domain expression matches the given email domain
+ *
+ * @param  {String}     emailDomain         The email domain to check
+ * @param  {String}     emailDomainExpr     The email domain expression (e.g., *.cam.ac.uk)
```
simong: I haven't gone through everything yet, but this example domain expression won't match most of Cambridge their email addresses. For example, my email address was `simon@cam.ac.uk`. Running `*.cam.ac.uk` against this email would not result in a match. `*cam.ac.uk` would however.

node_modules/oae-tenants/tests/test-tenants.js
```diff
@@ -218,6 +241,65 @@ describe('Tenants', function() {
...
+            var exactMatchTenantAlias = TenantsTestUtil.generateTestTenantAlias();
+            var exactMatchTenantDescription = TestsUtil.generateRandomText();
+            var exactMatchTenantHost = TenantsTestUtil.generateTestTenantHost();
+            var exactMatchTenantOpts = {'emailDomain': TenantsTestUtil.generateTestTenantHost()};
+
+            var suffixTenantAlias = TenantsTestUtil.generateTestTenantAlias();
+            var suffixTenantDescription = TestsUtil.generateRandomText();
+            var suffixTenantHost = TenantsTestUtil.generateTestTenantHost();
+            var suffixTenantOpts = {'emailDomain': util.format('*.%s', exactMatchTenantOpts.emailDomain)};
+
```
simong: It's a bit confusing that you're reusing the `exactMatchTenantOpts.emailDomain` here.

node_modules/oae-tenants/tests/test-tenants.js
```diff
@@ -218,6 +241,65 @@ describe('Tenants', function() {
         });
 
         /**
+         * Test that verifies a tenant can be looked up by an email domain match
+         */
+        it('verify get tenant by email domain', function(callback) {
+            var exactMatchTenantAlias = TenantsTestUtil.generateTestTenantAlias();
+            var exactMatchTenantDescription = TestsUtil.generateRandomText();
+            var exactMatchTenantHost = TenantsTestUtil.generateTestTenantHost();
+            var exactMatchTenantOpts = {'emailDomain': TenantsTestUtil.generateTestTenantHost()};
```
simong: Can we use `exactMatchTenantHost` as the `emailDomain` value as well? That makes it slightly easier to look at logs/debug values.

node_modules/oae-tenants/tests/test-tenants.js
simong: Can we try and use some real values here?

For example:
 * cam.ac.uk
 * admin.cam.ac.uk
 * sports.cam.ac.uk
 * uis.cam.ac.uk
 *  ...

should all resolve to the same tenant configured with `*cam.ac.uk`

node_modules/oae-tenants/config/config.js
simong: I was more thinking along the lines of the ConfigAPI trying to map data from the DB to fields that no longer exist

simong:
Couple of questions/follow-up remaining. Back to @mrvisser for review

node_modules/oae-tenants/lib/api.js
mrvisser: That's correct

mrvisser: Do you think there will be cases where a non-dotted suffix won't suffice?

E.g., if there is a `*cam.ac.uk`, maybe there would be a `ucam.ac.uk` that is a completely differen't college

If so, we might want to consider either:

* Support comma-separated domains: `cam.ac.uk,*.cam.ac.uk`
* Have `*.cam.ac.uk` match both exactly `cam.ac.uk` and `*.cam.ac.uk` suffix matches

comma-separated is more explicit and flexible, but more verbose to configure. Both allow for appropriate conflict detection, though.

simong: Could we try and "parse" the email domains and look for suffix matches? For example,
we configure `cam.ac.uk` as the Cambridge tenant's email domain and we want to check: `simon@admin.cam.ac.uk`, we explode all the email domain suffices:

1. uk
2. ac.uk
3. cam.ac.uk
4. admin.cam.ac.uk

and do _.contains(emailDomainSuffixes, emailDomain);

Something like this:
```

var getEmailSuffices = function(emailDomain) {
    var splittedEmailDomain = emailDomain.split('.');
    var emailSuffices = [];
    for (var i = 0; i < splittedEmailDomain.length; i++) {
       var emailSuffix = [];
       for (var j = i; j < splittedEmailDomain.length; j++) {
           emailSuffix.push(splittedEmailDomain[j]);
       }
       emailSuffices.push(emailSuffix.join('.'));
    }
    return emailSuffices;
};

var checkEmail = function(configuredEmailDomain, email) {
    var emailSuffices = getEmailSuffices(email.split('@')[1]);
    return _.contains(emailSuffices, configuredEmailDomain);
};

var test = function(email) {
    var configuredEmailDomain = 'cam.ac.uk';
    var containsDomain = checkEmail(configuredEmailDomain, email);
    console.log('%s contains %s: %s', email, configuredEmailDomain, containsDomain);
};

test('gaeremycks@gmail.com');
test('sg555@cam.ac.uk');
test('sg555@unicam.ac.uk');
test('support@uis.cam.ac.uk');
test('me@admin.foo.bar.cam.ac.uk');
```

mrvisser: I wasn't speaking in terms of the algorithm, but more the requirement. Given the implementation you posted, are you suggesting that `cam.ac.uk` implicitly support both `*.cam.ac.uk` and  the exact match of `cam.ac.uk` ?

If that's the case, it would be impossible to have a tenant that matches on both `cam.ac.uk` and then another tenant that matches on `someinterestgroup.cam.ac.uk`, for example.

If we allowed the user to put in `*.cam.ac.uk,cam.ac.uk`, then you have more flexibility for a tenant to match both the exact match and as a suffix. (e.g., one tenant with `cam.ac.uk`, and one tenant with `someinterestgroup.cam.ac.uk`.

simong: I'm not sure if that's something we need/want to support. /cc @nicolaasmatthijs 

If we do have to support that, we could try to find the best match by returning the tenant whose email domain is the longest?

mrvisser: I'm not sure doing smart things like using the longest match is feasible with 1000's of tenants, it seems just too tricky to manage.

I do think now we at least need to avoid having to do `*cam.ac.uk` as the suffix could potentially match `otherstuffcam.ac.uk`, and that could be switched to `*.cam.ac.uk,cam.ac.uk` to support a safer match.

mrvisser:
Followed up, back to @simong

simong:
LGTM. Can you add a note to [the upgrade guide](https://github.com/oaeproject/Hilary/wiki/OAE-Upgrade-Guide) about running the migration script? Feel free to merge this yourself once it's there.

mrvisser:
> LGTM. Can you add a note to the upgrade guide about running the migration script? Feel free to merge this yourself once it's there.

@simong this was done a while ago :) When it's merged, I'll move the note to "Next Release"

mrvisser:
Going to hold off merging this until after 11.3 as it is on its way out the door.

PR #1210, opened 2015-06-30
Pull passport-cas from npm

stuartf:
The SAML validation PR has been merged and published to npm so we can just pull it from there. As we don't really test CAS much in QA prior to releases, we should probably just bind to the fixed version?

### DIFF
--- a/package.json
+++ b/package.json
@@ -90,7 +90,7 @@
     "oauth2orize": "latest",
     "optimist": "latest",
     "passport": "latest",
-    "passport-cas": "git://github.com/oaeproject/passport-cas#samlValidateLoginUrl",
+    "passport-cas": "0.1.0",
     "passport-facebook": "git://github.com/oaeproject/passport-facebook",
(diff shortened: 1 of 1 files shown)


stuartf:
ready for review

package.json
```diff
@@ -90,7 +90,7 @@
     "oauth2orize": "latest",
     "optimist": "latest",
     "passport": "latest",
-    "passport-cas": "git://github.com/oaeproject/passport-cas#samlValidateLoginUrl",
+    "passport-cas": "0.1.0",
```
nicolaasmatthijs: Should this be latest ?

nicolaasmatthijs: @stuartf : Should this be `latest` ?

mrvisser: Discussed w/ Nico and we'll fix CAS since it's not really feasible get reassuring unit test coverage on it, so it's a big risk to let the version run away from us without supervision.

PR #1206, opened 2015-07-01
Split email from configuration into 2 separate fields

stuartf:
The email from configuration is currently a single field in the form of `Apereo OAE <noreply@research.unity.ac>`. This should be split into 2 separate config fields:

- Sender Name: this should default to the name of the tenant
- Sender Address: this should default to `noreply@<tenantdomain>`

### DIFF
--- a/node_modules/oae-email/lib/api.js
+++ b/node_modules/oae-email/lib/api.js
@@ -362,7 +362,9 @@ var sendEmail = module.exports.sendEmail = function(templateModule, templateId, 
 
     // If the `from` header is not set, we generate an intelligent `from` header based on the tenant host
     var tenant = TenantsAPI.getTenant(toUser.tenant.alias);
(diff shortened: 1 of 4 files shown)


node_modules/oae-email/lib/api.js
```diff
@@ -362,7 +362,9 @@ var sendEmail = module.exports.sendEmail = function(templateModule, templateId,
 
     // If the `from` header is not set, we generate an intelligent `from` header based on the tenant host
     var tenant = TenantsAPI.getTenant(toUser.tenant.alias);
-    var from = EmailConfig.getValue(tenant.alias, 'general', 'from') || util.format('"Apereo OAE" <noreply@%s>', tenant.host);
+    var fromName = EmailConfig.getValue(tenant.alias, 'general', 'fromName') || 'Apereo OAE';
```
nicolaasmatthijs: Note that this now defaults to `Apereo OAE`. However, it should default to the tenant displayname instead. @stuartf : Can you submit a follow-up PR for this?

PR #1205, opened 2015-07-07
Trim all form field values in requests

stuartf:
There are some bizarre characters in an import we've done: Tabs, trailing whitespace weirdly enough `\u000b`. These characters should be erradicated in the APIs when the tenant is created or updated.

---
This change is [Reviewable](https://reviewable.io/reviews/oaeproject/hilary/1205)

### DIFF
--- a/node_modules/oae-principals/tests/test-users.js
+++ b/node_modules/oae-principals/tests/test-users.js
@@ -1949,7 +1949,7 @@ describe('Users', function() {
                                             // Create user with an empty displayName
                                             RestAPI.User.updateUser(updateUserRestContext, userObj.id, {'displayName': '\n'}, function(err) {
                                                 assert.ok(err);
-                                                assert.equal(err.code, 401);
+                                                assert.equal(err.code, 400);
 
                                                 // Verify an incorrect visibility is invalid
                                                 RestAPI.User.updateUser(updateUserRestContext, userObj.id, {'visibility': 'so incorrect'}, function(err) {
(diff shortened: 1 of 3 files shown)


stuartf:
ready for review

node_modules/oae-principals/tests/test-users.js
```diff
@@ -1949,7 +1949,7 @@ describe('Users', function() {
                                             // Create user with an empty displayName
                                             RestAPI.User.updateUser(updateUserRestContext, userObj.id, {'displayName': '\n'}, function(err) {
                                                 assert.ok(err);
-                                                assert.equal(err.code, 400);
+                                                assert.equal(err.code, 401);
```
mrvisser: Why should this be a 401 now? Isn't it a user update missing a display name, which should still be 400?

node_modules/oae-util/lib/server.js
```diff
@@ -86,6 +86,16 @@ var setupServer = module.exports.setupServer = function(port, _config) {
...
 
+    // Trim whitespace on all post parameters
+    app.use(function(req, res, next) {
+        _.each(req.body, function(value, key) {
+            if (_.isString(value)){
+                req.body[key] = value.trim();
+            }
+        });
+        return next();
+    });
```
mrvisser: Can you write some tests for this?

stuartf: So, my first thought was to have the tests register a new route that would just echo back the body after it goes through all the middleware, then I could post a bunch of fields and verify that they all transform properly. That doesn't appear to work though, I think because there's a finalize step for the routes that runs before the tests. So, I'm open to suggestions on better ways to test this (or getting the route to register from a test).

simong: Is there anything wrong with trying to create a link that has trailing spaces in its displayName, description, user list, .. and then verifying they're not there once the item is created?

nicolaasmatthijs: Agreed. Lets just take one or a few of our endpoints and feed it some data with trailing spaces (including spaces only strings) and verify that it processes them as expected

sallakarppinen:
Seems to have conflicts that need resolving before this can be merged.

PR #1216, opened 2015-07-21
Activity summary emails are sent to deleted users

mrvisser:
The example I have observed is from a user on the *Unity implementation, the user was deleted on 13th July however they were still sent a weekly summary email on 17th July.

I can provide further details of the case but do not believe there to be anything unusual about this account, Therefore I believe this to be a general issue.

### DIFF
--- a/node_modules/oae-activity/lib/internal/email.js
+++ b/node_modules/oae-activity/lib/internal/email.js
@@ -91,7 +91,7 @@ ActivityEmitter.on(ActivityConstants.events.DELIVERED_ACTIVITIES, function(deliv
     scheduledEmailsCounter.incr();
 
     // We need to know the email preferences for each user so we know whether and when to send them an e-mail
-    PrincipalsDAO.getPrincipals(usersToMail, ['principalId', 'tenantAlias', 'email', 'emailPreference'], function(err, users) {
+    PrincipalsDAO.getPrincipals(usersToMail, ['principalId', 'tenantAlias', 'email', 'emailPreference', 'deleted'], function(err, users) {
         if (err) {
             scheduledEmailsCounter.decr();
             return log().error({'err': err, 'usersToMail': usersToMail}, 'Failed to get the email preference field for all the users in this activity');
@@ -99,7 +99,7 @@ ActivityEmitter.on(ActivityConstants.events.DELIVERED_ACTIVITIES, function(deliv
 
         var usersToQueue = _.filter(users, function(user) {
             // Users without a valid email address or who've opted out of email notifications should not get emails
-            return (user.email && user.emailPreference !== PrincipalsConstants.emailPreferences.NEVER);
+            return (!user.deleted && user.email && user.emailPreference !== PrincipalsConstants.emailPreferences.NEVER);
         });
 
         var emailBuckets = {};
--- a/node_modules/oae-activity/lib/internal/email.js
+++ b/node_modules/oae-activity/lib/internal/email.js
@@ -97,8 +97,8 @@ ActivityEmitter.on(ActivityConstants.events.DELIVERED_ACTIVITIES, function(deliv
             return log().error({'err': err, 'usersToMail': usersToMail}, 'Failed to get the email preference field for all the users in this activity');
         }
 
+        // Filter out users who should not get emails
         var usersToQueue = _.filter(users, function(user) {
-            // Users without a valid email address or who've opted out of email notifications should not get emails
             return (!user.deleted && user.email && user.emailPreference !== PrincipalsConstants.emailPreferences.NEVER);
         });
 
--- a/node_modules/oae-activity/tests/test-email.js
+++ b/node_modules/oae-activity/tests/test-email.js
@@ -883,7 +883,22 @@ describe('Activity Email', function() {
                                 // ensuring no emails actually get delivered
                                 EmailTestUtil.collectAndFetchEmailsForBucket(0, 'immediate', null, null, function(messages) {
                                     assert.strictEqual(messages.length, 0);
-                                    return callback();
+
+                                    // Restore mrvisser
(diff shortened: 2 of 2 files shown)


mrvisser:
Just figured it would be good to restore the user and then ensure a summary email gets sent in the test. Hold off on reviewing this.

mrvisser:
Fixeroo, assigned to @simong for review

mrvisser:
@simong I think we should prioritize this review before other larger features as it is most likely annoying to users who have deleted accounts in production.

node_modules/oae-activity/tests/test-email.js
```diff
@@ -855,6 +857,59 @@ describe('Activity Email', function() {
...
+                                    PrincipalsTestUtil.assertRestoreUserSucceeds(camAdminRestContext, mrvisser.user.id, function() {
+
+                                        // Simong will trigger another activity, annoying-sauce
+                                        RestAPI.Content.createLink(simong.restContext, 'Google', 'Awesome Google', 'public', 'http://www.google.ca', [mrvisser.user.id], [], [], function(err, link) {
+                                            assert.ok(!err);
+
+                                            // Collect and ensure mrvisser now gets the email for the latest activity
+                                            EmailTestUtil.collectAndFetchEmailsForBucket(0, 'immediate', null, null, function(messages) {
+                                                assert.strictEqual(messages.length, 1);
+                                                assert.strictEqual(messages[0].to[0].address, mrvisser.user.email);
```
simong: Can you verify that the email contains the last link and not the initial one

node_modules/oae-activity/lib/internal/email.js
```diff
@@ -99,7 +99,7 @@ ActivityEmitter.on(ActivityConstants.events.DELIVERED_ACTIVITIES, function(deliv
 
         var usersToQueue = _.filter(users, function(user) {
             // Users without a valid email address or who've opted out of email notifications should not get emails
```
simong: This comment is now out of sync

node_modules/oae-activity/tests/test-email.js
```diff
@@ -855,6 +857,59 @@ describe('Activity Email', function() {
     });
 
     /**
+     * Test that verifies that activity emails are not delivered to users who have deleted their
+     * profiles
+     */
+    it('verify email is not delivered to users who have since deleted their profile', function(callback) {
```
simong: What needs to happen to groups who have been deleted? Should we stop routing activities to them? /cc @nicolaasmatthijs

simong:
LGTM, the travis failure seems unrelated (although worrying). Back to @mrvisser for follow-up / answers

node_modules/oae-activity/tests/test-email.js
nicolaasmatthijs: Yeah, we should stop routing activities to deleted groups as well

mrvisser: Somehow I feel like we said we weren't going to do this for the group delete functionality because of performance drawbacks. Same with permissions.

@nicolaasmatthijs Should I just go ahead and just implement both of those?

mrvisser: Also where would the priority lay with respect to inviting guests? The work is not trivial and I'm assuming needs performance testing.

node_modules/oae-activity/tests/test-email.js
mrvisser: It will contain the initial one because it aggregates in the feed. But I can make sure it contains both.

mrvisser:
Followed up, back to @simong

node_modules/oae-activity/tests/test-email.js
nicolaasmatthijs: Ideally, a user would not receive any activities or email notifications through a deleted group being a member of an active group or being a member of a content item. Can you describe the potential performance drawbacks that are involved here?

Regardless, priority of this is definitely lower than inviting guests.

mrvisser: > Can you describe the potential performance drawbacks that are involved here?

The draw back is that every time you expand the authz memberships of a group you need to run the ids through `AuthzDeleted.isDeleted` (cassandra multi-get). So that would be permissions checks, activity routing, etc...

davidoae:
Just to be sure you know, the user is still getting the summary emails.
This is an information exposure bug. In theory an administrator could create a user, add to lots of groups, then delete the user and the email recipient would be able to read content without anyone knowing.

nicolaasmatthijs:
@davidoae : This will be rolled out as part of the v12 release sometime this week.

mrvisser:
Hi David, that's correct, a release has not been done for this bug yet so it wouldn't be witnessed in production.

> This is an information exposure bug. In theory an administrator could create a user, add to lots of groups, then delete the user and the email recipient would be able to read content without anyone knowing.

I respectfully disagree that it's an information exposure bug. The manager of each group can still see the deleted user's entry in the members list of the group, and has control to remove the deleted user entry from the group if they wish. The manager(s) also received a notification when the user was added, so they can then decide how to handle the fact that this user was given access at the time that it happened. The manager(s) of the group, however, are not given a notification when the user account was deleted, so the whole process of granting access/deleting is transparent to everyone except the administrator.

The reason for it to work in this way is that, since a user's account can be restored along with their group memberships, it's important to still give group managers the ability to make decisions on deleted user access.

yojimbo23:
So it sounds like the easiest immediate fix to the problem is simply to remove the user from the groups that are causing him to get notifications.

Am I right in thinking that there is no way to view or manage deleted users via the admin interface?

mrvisser:
Yes I think that's the quickest way to stop the emails. There is currently no UI for managing deleted users. There are back-end REST endpoints available for searching and restoring users, but that's probably not very helpful.

yojimbo23:
That's fine for now imo. Potential extensions to admin functionality are something we should discuss elsewhere. I am making a few notes on this.

PR #1199, opened 2015-07-22
Report basic stats in admin UI

Coenego:
The admin UI for a tenancy should report the following basic stats on the Tenant page:
- The number of (unique) active users in the current month. An active users is defined as user that has either signed in or checked its activity feed.
- The total file size uploaded in the current month.

This data should be readily available in MixPanel.

---
This change is [Reviewable](https://reviewable.io/reviews/oaeproject/hilary/1199)


Coenego:
Still WIP. Assigning to @mrvisser for a quick look.

mrvisser:
Thanks @Coenego . As discussed I'll have a look this week.

config.js
```diff
@@ -447,5 +447,7 @@ config.tincanapi = {
  */
 config.mixpanel = {
     'enabled': false,
+    'apikey': '',
+    'apisecret': '',
     'token': 'f3e9fce119d357b745a8dfa36248d632'
```
mrvisser: If we can use the token to do API requests we should, but I'm guessing it's not since you added these.

node_modules/oae-mixpanel/lib/api.js
```diff
@@ -14,24 +14,132 @@
...
+var exporter = null;
 
 /**
  * Initialize mixpanel event tracking
  *
  * @param  {Object}     config      An object containing the full system configuration (i.e., `config.js`)
  */
-var init = module.exports.init = function(config) {
+var init = module.exports.init = function(_config) {
+    config = _config;
```
mrvisser: Handling both `_config` and `config` doesn't seem to be necessary

node_modules/oae-mixpanel/lib/api.js
```diff
@@ -14,24 +14,132 @@
...
+ *
+ * @param  {Context}    ctx                 Standard context object containing the current user and the current tenant
+ * @param  {Function}   callback            Standard callback function
+ * @param  {Object}     callback.err        An error that occurred, if any
+ * @param  {Object}     callback.stats      Object containing the stats for the tenant(s)
+ */
+var getUniqueUsers = module.exports.getUniqueUsers = function(ctx, callback) {
+    if (ctx.user() && !ctx.user().isAdmin()) {
+        return callback({'code': 401, 'msg': 'Only administrators can request stats'});
+    }
```
mrvisser: This logic allows anonymous users to request stats

node_modules/oae-mixpanel/lib/api.js
```diff
@@ -14,24 +14,132 @@
...
+    var data = {
+        'from_date': _getStartDate(Date.now()),
+        'to_date': moment.utc(Date.now()).format('YYYY-MM-DD'),
+        'event': 'userLoggedOut',
+        'on': 'properties["tenant"]',
+        'type': 'unique'
+    };
+
+    // Return the unique users for a specific tenant if the endpoint was accessed on the tenant router
+    if (ctx.tenant().alias !== 'admin') {
```
mrvisser: `ctx.tenant().isGlobalAdminServer` can be used instead of checking the tenant alias

node_modules/oae-mixpanel/lib/api.js
```diff
@@ -14,24 +14,132 @@
...
+        data = _.extend(data, {
```
mrvisser: Since `_.extend` will mutate `data`, we don't have to re-assign here with `data =`

node_modules/oae-mixpanel/lib/api.js
```diff
@@ -14,24 +14,132 @@
...
+            'where': '"' + ctx.tenant().alias + '" == properties["tenant"]'
```
mrvisser: We try and use `util.format` for string when templating strings:

```javascript
util.format('"%s" == properties["tenant"]', ctx.tenant().alias);
```

node_modules/oae-mixpanel/lib/api.js
```diff
@@ -14,24 +14,132 @@
...
+ * @param  {Context}    ctx                 Standard context object containing the current user and the current tenant
+ * @param  {String}     alias               The alias of the tenant to get the stats for
+ * @param  {Function}   callback            Standard callback function
+ * @param  {Object}     callback.err        An error that occurred, if any
+ * @param  {Object}     callback.stats      Object containing the stats for the tenant
+ */
+var getUniqueUsersForTenant = module.exports.getUniqueUsersForTenant = function(ctx, alias, callback) {
+    if (ctx.user() && !ctx.user().isAdmin()) {
+        return callback({'code': 401, 'msg': 'Only administrators can request stats'});
+    }
```
mrvisser: Lets anonymous use the endpoint

node_modules/oae-mixpanel/lib/api.js
```diff
@@ -14,24 +14,132 @@
...
+
+    // Request options object
+    var data = {
+        'from_date': _getStartDate(Date.now()),
+        'to_date': moment.utc(Date.now()).format('YYYY-MM-DD'),
+        'event': 'userLoggedIn',
+        'on': 'properties["tenant"]',
+        'where': '"' + alias + '" == properties["tenant"]',
+        'type': 'unique'
+    };
```
mrvisser: You can use `util.format` for the `'where'` key. Also, we should try and separate out the similar parts of building this `data` object so we aren't repeating it in multiple places

node_modules/oae-mixpanel/lib/api.js
```diff
@@ -14,24 +14,132 @@
...
+
+/**
+ * Return the start date
+ *
+ * @param  {Date}       [date]      The current date
+ * @return {String}                 The start date in ISO 8601 format (e.g. 2015-07-01)
+ * @api private
+ */
+var _getStartDate = function(date) {
+    date = date || Date.now();
```
mrvisser: It looks like all the handling for "date" has been millis since epoch. Can rename `date` to `startMillis`?

Also, all places where we do `_getStartDate(Date.now());`, we can just change to `_getStartDate()` since it's the default.

node_modules/oae-mixpanel/lib/api.js
```diff
@@ -14,24 +14,132 @@
...
+    var day = parseInt(moment(date).date(), 10);
```
mrvisser: `moment(date).date()` will already return a number. What are you trying to handle here?

node_modules/oae-mixpanel/lib/api.js
```diff
@@ -14,24 +14,132 @@
...
+    return moment(date).subtract({'days': day - 1}).utc().format('YYYY-MM-DD');
+};
```
mrvisser: I think this method could be rewritten as:

```javascript
startMillis = startMillis || Date.now();
return moment(startMillis).date(1).utc().format('YYYY-MM-DD');
```

node_modules/oae-mixpanel/lib/rest.js
```diff
@@ -0,0 +1,86 @@
...
+ */
+OAE.tenantRouter.on('get', '/api/mixpanel/stats/uniqueUsers', function(req, res) {
+    MixPanelAPI.getUniqueUsers(req.ctx, function(err, data) {
+        if (err) {
+            return res.status(err.code).send(err.msg);
+        }
+
+        return res.status(200).send(data);
+    });
+});
```
mrvisser: I think this second `/api/mixpanel/stats/uniqueUsers` is unnecessary.

mrvisser:
The approach here looks fine, so I did some lower level review. Let me know when you'd like another round of review @Coenego

Coenego:
Back to @mrvisser for review after follow-up.

I kept the `apikey` and `apisecret` in the config.js file since the `mixpanel-data-export` module requires these instead of the `token`.

config.js
Coenego: Unfortunately these need to be kept here since the `mixpanel-data-export` module uses them to create the signature for the request.

node_modules/oae-mixpanel/lib/api.content.js
```diff
@@ -0,0 +1,124 @@
...
+ *  * Exporter return example:
+ *  *
+ *  *   {
+ *  *       status: "ok",
+ *  *       results: {
+ *  *           2015-07-29: 2489,
+ *  *           2015-07-30: 2989176,
+ *  *           2015-07-31: 670234
+ *  *       }
+ *  *   }
```
mrvisser: The extra `*`'s being used here are markdown bullet lists, not code. I think it should instead be:

    * Exporter return example
    *
    * ```
    * {
    *     "status": "ok",
    *     ...
    * }
    * ```

node_modules/oae-mixpanel/lib/api.content.js
```diff
@@ -0,0 +1,124 @@
...
+ */
+
+var _ = require('underscore');
+var Validator = require('oae-util/lib/validator').Validator;
+
+var MixPanelUtil = require('./util');
+
+/**
+ * Get the total size of uploaded files for the current month for all tenants
+ * or the current tenant, depending on the used router.
```
mrvisser: `depending on the used router`

The documentation can be in terms of what tenant is being used rather than a "router". So you could say instead "depending on if the request is performed on the global admin tenant or a user tenant"

node_modules/oae-mixpanel/lib/api.content.js
```diff
@@ -0,0 +1,124 @@
...
+ *  *       }
+ *  *   }
+ *
+ * @param  {Context}    ctx                 Standard context object containing the current user and the current tenant
+ * @param  {Function}   callback            Standard callback function
+ * @param  {Object}     callback.err        An error that occurred, if any
+ * @param  {Object}     callback.stats      Object containing the stats for the tenant(s)
+ */
+var getUploadedFilesFileSize = module.exports.getUploadedFilesFileSize = function(ctx, callback) {
+    var mixPanelConfig = require('../../../config').config.mixpanel;
```
mrvisser: The application configuration is not obtained in this way, because a different config file location can be used.

Instead, you should hook into the initialize sequence with an `init.js` and initialize this module with the configuration with which the application is being initialized. See other `init.js` files for reference.

node_modules/oae-mixpanel/lib/api.content.js
```diff
@@ -0,0 +1,124 @@
...
+    // Return the unique users for a specific tenant if the endpoint was accessed on the tenant router
+    var alias = null;
+    if (!ctx.tenant().isGlobalAdminServer) {
+        alias = ctx.tenant().alias;
+    }
+
+    // Request options object
+    var options = MixPanelUtil.createRequestOptionsObject('createdContent', alias, {
+        'on': 'properties["filesize"]'
+    });
```
mrvisser: Can we better classify the naming of this object to indicate it's creating an exporter request options? It could perhaps be:

`MixPanelUtil.createExporterOptions(...)`

node_modules/oae-mixpanel/lib/api.js
```diff
@@ -16,8 +16,11 @@
 var _ = require('underscore');
 var Mixpanel = require('mixpanel');
 
+var ContentAPI = require('./api.content');
+var UsersAPI = require('./api.users');
```
mrvisser: Let's call these `MixPanelContentAPI` and `MixPanelUsersAPI`, just so they aren't confused with wide-spread familiarity of these being the actual `ContentAPI` and `UserAPI`

node_modules/oae-mixpanel/tests/test-content.js
```diff
@@ -0,0 +1,162 @@
...
+     */
+    beforeEach(function() {
+        require('../../../config').config.mixpanel.enabled = true;
+    });
+
+    /**
+     * Disable MixPanel when all the tests are ran
+     */
+    after(function() {
+        require('../../../config').config.mixpanel.enabled = false;
```
mrvisser: When switching to the application initialization method of handling config, you could instead put functions on the MixPanelAPI that allows you to enable and disabled it.

node_modules/oae-mixpanel/tests/test-content.js
```diff
@@ -0,0 +1,162 @@
...
+    });
+
+    describe('#getUploadedFilesFileSize()', function() {
+
+        /**
+         * Test that verifies that an error is thrown when stats are requested and MixPanel is disabled
+         */
+        it('Verify that an error is thrown when stats are requested and MixPanel is disabled', function(callback) {
```
mrvisser: The `it` messages do not start with a capital

node_modules/oae-mixpanel/tests/test-users.js
```diff
@@ -0,0 +1,162 @@
...
+
+var _ = require('underscore');
+var assert = require('assert');
+
+var RestAPI = require('oae-rest');
+var TestsUtil = require('oae-tests');
+
+var MixPanelTestUtil = require('oae-mixpanel/lib/test/util');
+
+describe('MixPanel - Users', function() {
```
mrvisser: Rather than scoping with text conventions like a dash, we can instead use nested `describe` blocks:

```javascript
describe('MixPanel', function() {
    describe('Users', function() {
```

node_modules/oae-mixpanel/lib/api.content.js
mrvisser: Actually, can this function not take in an `alias` like the other ones do?

mrvisser: I see what's going on. Let's alter the API layout a bit if possible:

1. Don't have variations for `ForTenant` and having a tenant
2. Have all stats methods take in a `tenantAlias`
3. If `alias` is not specified, default it to the current tenant
4. All other logic is the same thereon (the current user in context must be able to administer the specified tenant, etc...)

mrvisser:
Finished another round of review, back to @Coenego

Coenego:
Back to @mrvisser for review after follow-up

config.js
```diff
@@ -275,7 +275,7 @@ config.mq = {
  * @param  {String}      credentials.password           The password to login with on the global admin server
  */
 config.previews = {
-    'enabled': false,
+    'enabled': true,
```
sallakarppinen: We do want to keep previews disabled by default.

config.js
```diff
@@ -446,6 +446,8 @@ config.tincanapi = {
  * @param  {String}     token           The mixpanel api token
  */
 config.mixpanel = {
-    'enabled': false,
-    'token': 'f3e9fce119d357b745a8dfa36248d632'
+    'enabled': true,
```
sallakarppinen: And we should probably keep MixPanel disabled by default as well, plus we should not add real values as the `apikey` and `apisecret` for a paid service.

package.json
```diff
@@ -78,7 +80,7 @@
     "oae-mixpanel": "",
     "oae-preview-processor": "",
     "oae-principals": "",
-    "oae-rest": "latest",
+    "oae-rest": "git://github.com/Coenego/oae-rest#mixpanel-rest-api",
```
sallakarppinen: The direct path needs to be taken out and any necessary PRs merged to the main oae-rest repo.

davidoae:
I think total disk size used by the tenancy should also be reported.
The information cannot be determined on the fly but I can store the information daily in any way you wish.

PR #1195, opened 2015-07-28
Tenant list should be paged and searchable

stuartf:
We will soon have 1000+ tenants in the production environment. The list of available tenancies in the admin UI should become (infinitely) paged and we should create a separate endpoint to search through the list of available tenancies.

After this, we should run a performance test with 1000s of tenancies (and their users) to ensure that having this number of tenancies doesn't cause any performance regressions.

### DIFF
--- a/node_modules/oae-tenants/lib/api.js
+++ b/node_modules/oae-tenants/lib/api.js
@@ -65,14 +65,21 @@ Pubsub.on('oae-tenants', function (message) {
     var cmd = args.shift();
     var alias = args.shift();
 
-    // Re-cache the available tenants
-    _cacheTenants(function(err) {
+    var callback = function(err) {
         if (err) {
             return log().error('There was an error caching the tenants');
         }
 
         TenantsAPI.emit(cmd, getTenant(alias));
-    });
+    };
+
+    if (_.contains(['start', 'stop', 'created'], cmd)) {
+        // Update the cached version of the tenant
+        _updateCachedTenant(alias, callback);
+    } else {
+        // Re-cache the available tenants
+        _cacheTenants(callback);
+    }
 });
 
 /**
@@ -225,6 +232,52 @@ var _cacheTenants = function(callback) {
 
         // Cache the sorted list
         sortedTenants = _.sortBy(tenants, 'alias');
+
+        // Indicate that all tenants have been cached
+        TenantsAPI.emit('cached');
+
+        return callback(null, tenants);
+    });
+};
+
+/**
+ * Fetch the tenant and update it's entry in the cache
+ *
+ * @param  {String}         tenantAlias         The alias of the tenant to be re-cached
+ * @param  {Function}       [callback]          Standard callback function
+ * @param  {Object}         [callback.err]      Error object containing the error code and message
+ * @api private
+ */
+var _updateCachedTenant = function(tenantAlias, callback) {
+    callback = callback || function(err) {
+        if (err) {
+            log().error({'err': err}, 'Failed to re-cache the tenants');
+        }
+    };
+
+    // Get the available tenants
+    Cassandra.runQuery('SELECT * FROM "Tenant" WHERE "alias" = ?', [tenantAlias], function(err, rows) {
+        if (err) {
+            return callback(err);
+        }
+
+        _.each(rows, function(row) {
+            var tenant = mapToTenant(row);
+            // Cache all tenants
+            tenants[tenant.alias] = tenant;
+            tenantsByHost[tenant.host] = tenant;
+
+            // Insert at the correct location in the sorted list
+            var index = _.findIndex(sortedTenants, function(t){
+                return t.alias === tenantAlias;
+            });
+            if (index === -1) {
+                index = _.sortedIndex(sortedTenants, tenant, 'alias');
+                sortedTenants.splice(index, 0, tenant);
+            } else {
+                sortedTenants[index] = tenant;
+            }
+        });
 
         // Indicate that all tenants have been cached
         TenantsAPI.emit('cached');
--- a/node_modules/oae-tenants/lib/api.js
+++ b/node_modules/oae-tenants/lib/api.js
@@ -35,6 +35,8 @@ var tenants = {};
 // Variable that caches the available tenants, keyed by their hostname. This will be used for
 // quick look-ups when checking whether or not a hostname is associated to a tenant
 var tenantsByHost = {};
+// Variable that caches the available tenants sorted by alias
+var sortedTenants = [];
 
 // Variable that caches the global administration tenant object
 var globalTenant = null;
@@ -145,8 +147,7 @@ var getTenants = module.exports.getTenants = function(excludeDisabled, start, li
 
     var filteredTenants = {};
     // Filter out the admin tenant and disabled tenants if requested
-    var preFiltered = excludeDisabled? _.where(tenants, {active: true, isGlobalAdminServer: false}): _.where(tenants, {isGlobalAdminServer: false});
-    preFiltered = _.sortBy(preFiltered, 'alias');
+    var preFiltered = excludeDisabled? _.where(sortedTenants, {active: true, isGlobalAdminServer: false}): _.where(sortedTenants, {isGlobalAdminServer: false});
     // Calculate the end index for paging
     var end = preFiltered.length;
     if (limit) {
@@ -215,6 +216,9 @@ var _cacheTenants = function(callback) {
             tenants[tenant.alias] = tenant;
             tenantsByHost[tenant.host] = tenant;
         });
+
+        // Cache the sorted list
+        sortedTenants = _.sortBy(tenants, 'alias');
 
         // Indicate that all tenants have been cached
         TenantsAPI.emit('cached');
--- a/node_modules/oae-tenants/lib/api.js
+++ b/node_modules/oae-tenants/lib/api.js
@@ -132,15 +132,23 @@ var init = module.exports.init = function(_serverConfig, callback) {
  * Get a list of all available tenants from cache. The global admin tenant will be excluded from the resulting tenant list
  *
  * @param  {Boolean}        [excludeDisabled]   Whether or not disabled tenants should be included. By default, all tenants will be returned
+ * @param  {Number}         [start]             The index to start paging at, defaults to 0
+ * @param  {Number}         [limit]             The maximum number of tenants to return, if not specified all tenants will be returned
  * @return {Object}                             An object keyed by tenant alias holding all the tenants
  */
-var getTenants = module.exports.getTenants = function(excludeDisabled) {
+var getTenants = module.exports.getTenants = function(excludeDisabled, start, limit) {
     var filteredTenants = {};
-    _.each(tenants, function(tenant, tenantAlias) {
-        // Exclude all disabled tenants when `exludeDisabled` has been provided
-        if (!tenant.isGlobalAdminServer && (!excludeDisabled || tenant.active)) {
-            filteredTenants[tenantAlias] = _copyTenant(tenant);
-        }
+    start = start || 0;
+    // Filter out the admin tenant and disabled tenants if requested
+    var preFiltered = excludeDisabled? _.where(tenants, {active: true, isGlobalAdminServer: false}): _.where(tenants, {isGlobalAdminServer: false});
+    preFiltered = _.sortBy(preFiltered, 'alias');
+    // Calculate the end index for paging
+    var end = preFiltered.length;
+    if (limit) {
+        end = parseInt(limit) + parseInt(start);
+    }
+    _.each(preFiltered.slice(start, end), function(tenant) {
+        filteredTenants[tenant.alias] = _copyTenant(tenant);
     });
     return filteredTenants;
 };
--- a/node_modules/oae-tenants/lib/api.js
+++ b/node_modules/oae-tenants/lib/api.js
@@ -14,6 +14,7 @@
  */
 
 var _ = require('underscore');
+var lunr = require('lunr');
 var util = require('util');
 
 var Cassandra = require('oae-util/lib/cassandra');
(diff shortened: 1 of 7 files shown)


davidoae:
Stuart, FYI we expect to import >20k tenancies, the performance test should probably reflect this kind of number.

stuartf:
Ready for review. The travis failures are due to needing the patched oae-rest

simong:
Have you tried creating 20k tenants and see how this impact things like search? We seem to get all the aliases you can interact with and then throw that at elasticsearch. At the very least I would expect some performance degradations when sending a search request with 20k aliases in it.

node_modules/oae-tenants/lib/api.js
```diff
@@ -132,15 +132,23 @@ var init = module.exports.init = function(_serverConfig, callback) {
...
     var filteredTenants = {};
-    _.each(tenants, function(tenant, tenantAlias) {
-        // Exclude all disabled tenants when `exludeDisabled` has been provided
-        if (!tenant.isGlobalAdminServer && (!excludeDisabled || tenant.active)) {
-            filteredTenants[tenantAlias] = _copyTenant(tenant);
-        }
+    start = start || 0;
+    // Filter out the admin tenant and disabled tenants if requested
+    var preFiltered = excludeDisabled? _.where(tenants, {active: true, isGlobalAdminServer: false}): _.where(tenants, {isGlobalAdminServer: false});
+    preFiltered = _.sortBy(preFiltered, 'alias');
```
simong: As we're going to have 20k objects in here, it might make sense to have these cached and pre-sorted somewhere?

node_modules/oae-tenants/lib/rest.js
```diff
@@ -219,11 +219,13 @@ OAE.globalAdminRouter.on('post', '/api/tenant/create', function(req, res) {
  * @Server      admin
  * @Method      GET
  * @Path        /tenants
+ * @QueryParam  {number}        [start]         The index to start paging from
+ * @QueryParam  {number}        [limit]         The maximum number of results to return
```
simong: We should default these to 10

node_modules/oae-tenants/lib/rest.js
```diff
@@ -219,11 +219,13 @@ OAE.globalAdminRouter.on('post', '/api/tenant/create', function(req, res) {
...
  * @Return      {Tenants}                       All available tenants
  * @HttpResponse                200             Tenants available
  */
 OAE.globalAdminRouter.on('get', '/api/tenants', function(req, res, next) {
-    var tenants = TenantsAPI.getTenants();
+    var tenants = TenantsAPI.getTenants(false, req.query.start, req.query.limit);
```
simong: As we can't default the limit in the API, we should do that here:

var limit = OaeUtil.getNumberParam(req.query.limit, 10, 1);
var tenants = TenantsAPI.getTenants(false, req.query.start, limit);

node_modules/oae-tenants/lib/api.js
```diff
@@ -132,15 +132,23 @@ var init = module.exports.init = function(_serverConfig, callback) {
  * Get a list of all available tenants from cache. The global admin tenant will be excluded from the resulting tenant list
  *
  * @param  {Boolean}        [excludeDisabled]   Whether or not disabled tenants should be included. By default, all tenants will be returned
+ * @param  {Number}         [start]             The index to start paging at, defaults to 0
+ * @param  {Number}         [limit]             The maximum number of tenants to return, if not specified all tenants will be returned
  * @return {Object}                             An object keyed by tenant alias holding all the tenants
  */
-var getTenants = module.exports.getTenants = function(excludeDisabled) {
+var getTenants = module.exports.getTenants = function(excludeDisabled, start, limit) {
```
simong: We try to default the function parameters at the top of the function by using OaeUtil. So this could become:

```javascript
var getTenants = module.exports.getTenants = function(excludeDisabled, start, limit) {
    excludeDisabled = OaeUtil.castToBoolean(excludeDisabled);
    start = OaeUtil.getNumberParam(start, 0);
    // Note that we don't apply a default limit here as some methods need ALL tenants
    limit = OaeUtil.getNumberParam(limit);
```

node_modules/oae-tenants/tests/test-tenants.js
```diff
@@ -89,7 +87,12 @@ describe('Tenants', function() {
                         // Verify that the global admin tenant is not included
                         assert.ok(!tenants['admin']);
 
-                        callback();
+                        // Check that paging works
+                        RestAPI.Tenants.getTenants(globalAdminRestContext, 0, 1, function(err, tenants) {
```
simong: These tests will have to verify that doing `0, 1` gets the first tenant and `1, 1` gets the second tenant

simong:
Some follow-up and a question remains. Back to @stuartf

stuartf:
Ready for further review.

stuartf:
I've only been able to get ~15K tenants on my dev machine, but search still works with little delay

mrvisser:
A persistent slowdown in adding new tenants, even after restarting Hilary, after you have 15k of them seems like cause for concern. Have you figured out why adding many tenants causes things to slow down?

nicolaasmatthijs:
Is it because we re-cache the full list of tenants every time a new tenant is added?

mrvisser:
I thought we just pulled them up individually, but it looks like we do load them all. That's probably it.

node_modules/oae-tenants/lib/api.js
```diff
@@ -132,15 +135,26 @@ var init = module.exports.init = function(_serverConfig, callback) {
  * Get a list of all available tenants from cache. The global admin tenant will be excluded from the resulting tenant list
  *
  * @param  {Boolean}        [excludeDisabled]   Whether or not disabled tenants should be included. By default, all tenants will be returned
+ * @param  {Number}         [start]             The index to start paging at, defaults to 0
+ * @param  {Number}         [limit]             The maximum number of tenants to return, if not specified all tenants will be returned
```
nicolaasmatthijs: We should provide a way in which users can search through the list of tenants (matches displayname and host). I'm fine with this being added as a simple `q` parameter on top of the existing endpoint and I'm fine with this being a simple in memory search (if performant enough), but it has to be part of this iteration as an infinitely scrolling tenant list without search is going to be unusable.

If an in-memory search is too slow, we'll have to look at indexing the tenants in Elastic Search ...

nicolaasmatthijs:
Back to @stuartf for search functionality and an assessment of whether in-memory search is fast enough.

stuartf:
I've added the query parameter, but not addressed the re-caching issue yet.

stuartf:
ready for more review

node_modules/oae-tenants/lib/api.js
```diff
@@ -132,15 +142,32 @@ var init = module.exports.init = function(_serverConfig, callback) {
...
+ * @param  {String}         [q]                 Query string used to filter results
  * @return {Object}                             An object keyed by tenant alias holding all the tenants
  */
-var getTenants = module.exports.getTenants = function(excludeDisabled) {
+var getTenants = module.exports.getTenants = function(excludeDisabled, start, limit, q) {
```
mrvisser: Can you convert these 4 parameters into an `opts` parameter?

node_modules/oae-tenants/lib/api.js
```diff
@@ -132,15 +142,32 @@ var init = module.exports.init = function(_serverConfig, callback) {
...
+    limit = OaeUtil.getNumberParam(limit);
+
     var filteredTenants = {};
-    _.each(tenants, function(tenant, tenantAlias) {
-        // Exclude all disabled tenants when `exludeDisabled` has been provided
-        if (!tenant.isGlobalAdminServer && (!excludeDisabled || tenant.active)) {
-            filteredTenants[tenantAlias] = _copyTenant(tenant);
-        }
+    // Filter out the admin tenant and disabled tenants if requested
+    var preFiltered = excludeDisabled? _.where(sortedTenants, {active: true, isGlobalAdminServer: false}): _.where(sortedTenants, {isGlobalAdminServer: false});
```
mrvisser: This expression needs to be simplified. Something like:

```javascript
var matchedTenants = _.filter(sortedTenants, {'isGlobalAdminServer': false});
if (excludeDisabled) {
    matchedTenants = _.filter(matchedTenants, {'active': true});
}
...
}```

Comment (mrvisser) [inline node_modules/oae-tenants/lib/api.js]: Also object keys are always in quotes.

File: node_modules/oae-tenants/lib/api.js [hunk 9]
```diff
@@ -203,6 +230,55 @@ var _cacheTenants = function(callback) {
...
+            var index = _.findIndex(sortedTenants, function(t){
+                return t.alias === tenantAlias;
+            });
+            if (index === -1) {
+                index = _.sortedIndex(sortedTenants, tenant, 'alias');
+                sortedTenants.splice(index, 0, tenant);
+            } else {
+                sortedTenants[index] = tenant;
+            }
+        });
```
mrvisser: There will only be 1 row here, so representing as a foreach is misleading. It should just grab the first one, and warn if there aren't any

node_modules/oae-tenants/lib/api.js
```diff
@@ -132,15 +142,32 @@ var init = module.exports.init = function(_serverConfig, callback) {
...
-    _.each(tenants, function(tenant, tenantAlias) {
-        // Exclude all disabled tenants when `exludeDisabled` has been provided
-        if (!tenant.isGlobalAdminServer && (!excludeDisabled || tenant.active)) {
-            filteredTenants[tenantAlias] = _copyTenant(tenant);
-        }
+    // Filter out the admin tenant and disabled tenants if requested
+    var preFiltered = excludeDisabled? _.where(sortedTenants, {active: true, isGlobalAdminServer: false}): _.where(sortedTenants, {isGlobalAdminServer: false});
+    if (_.isString(q)) {
+        preFiltered = _.filter(preFiltered, function(tenant) {
+            return tenant.alias.indexOf(q) !== -1 || tenant.displayName.indexOf(q) !== -1 || tenant.host.indexOf(q) !== -1;
```
mrvisser: I feel like this search should be case insensitive.

mrvisser: I'm not sure strict `indexOf` even with substring will be good enough. @nicolaasmatthijs wdyt?

I think we should either pull in a library like `lunr.js` to do in-memory indexing, or hook it up to elasticsearch.

nicolaasmatthijs: If this search was restricted to just the global admin UI, the substring approach would have been sufficient. However, if this search will also be used for the WAYF screen, it will probably be insufficient. For example, searching for "Cambridge University" should return "University of Cambridge".

nicolaasmatthijs: I don't feel strong about lunr.js vs elasticsearch, as long as it perform well with 20k institutions.

mrvisser: @stuartf I'm working on pulling in `lunr.js` and I'll submit a PR to your PR

image

But that shouldn't block the performance testing setup, etc...

mrvisser:
@stuartf please note PR https://github.com/stuartf/Hilary/pull/3 to add tenant search indexing

node_modules/oae-tenants/lib/api.js
```diff
@@ -14,16 +14,19 @@
  */
 
 var _ = require('underscore');
+var lunr = require('lunr');
```
simong: Unused

node_modules/oae-tenants/lib/api.js
```diff
@@ -142,10 +159,71 @@ var getTenants = module.exports.getTenants = function(excludeDisabled) {
...
+
+    // Filter out the global admin server and deleted tenancies
+    results = _.filter(results, function(tenant) {
+        return (!tenant.isGlobalAdminServer);
+    });
+
+    // Filter out disabled or deleted tenants if they aren't said to be included
+    if (!includeDisabled) {
+        results = _.where(results, {'active': true, 'deleted': false});
+    }
```
simong: Can we combine the `isGlobalAdmin` and `includeDisabled` check so we only have to iterate over the potentially 5k array once?

```javascript
var filter = {'isGlobalAdminServer': false};
if (!includeDisabled) {
    filter.active = true;
    filter.deleted = false;
}
results = _.where(results, filter);
```

node_modules/oae-tenants/lib/api.js
```diff
@@ -203,6 +281,60 @@ var _cacheTenants = function(callback) {
...
+        }
+    };
+
+    // Get the available tenants
+    Cassandra.runQuery('SELECT * FROM "Tenant" WHERE "alias" = ?', [tenantAlias], function(err, rows) {
+        if (err) {
+            return callback(err);
+        }
+
+        _.each(rows, function(row) {
```
simong: There should only be 1 row as the `alias` field is the PK for this column family. I think we can get rid of this `_.each` here

simong:
LGTM, couple of super minor issues. Back to @stuartf for follow-up

stuartf:
fixed review issues

node_modules/oae-tenants/lib/api.js
mrvisser: I think I had only done this to handle `isGlobalAdminServer` safer with a `false`y check instead of strict `false` check. There shouldn't be any issue implementing this suggestion.

node_modules/oae-tenants/lib/internal/tenantIndex.js
```diff
@@ -0,0 +1,77 @@
+
```
mrvisser: Oops, needs license.

mrvisser:
Sorry, just one more

stuartf:
ready for more review

mrvisser:
Closing this in favour of a new PR coming in w/ some performance improvements based on testing this week.

PR #1227, opened 2015-09-04
(TenantPaging) Add APIs and performance improvements needed for supporting 20k+ tenancies

mrvisser:


node_modules/oae-authentication/lib/api.js
```diff
@@ -87,7 +87,6 @@ var _configUpdate = function(tenantAlias) {
     }
 };
 
-ConfigAPI.on('cleared', _configUpdate);
```
simong: Can we do this? I can't remember of the top of my head but are there some strategies that we configure globally?

mrvisser: That event never fires, `updated` fires instead

mrvisser: Also, what do some strategies being configured globally have to do with this particular event?

simong: I figured if a globally configured key gets cleared it might have an impact on all strategies.

mrvisser: I see what you mean, yea it could have an effect. The `update` event seems to have been handling that case, so this should be fine.

simong:
LGTM, this can be merged in once the UI is ready. Marking as On Hold and assigning to @nicolaasmatthijs

nicolaasmatthijs:
Merged

PR #1225, opened 2015-09-17
Store country for tenant

mrvisser:
We should store the country for each tenant. This country will be used for the "Where Are You From" autosuggest, as well as the tenancy picker on the OAE website.


mrvisser:
On hold until inviting guests is merged

node_modules/oae-activity/lib/internal/config.js
```diff
@@ -32,7 +32,7 @@ var DEFAULT_MAIL_POLLING_FREQUENCY = 600;           // 10 minutes (in seconds)
 var DEFAULT_MAIL_DAILY_HOUR = 0;                    // Midnight
 var DEFAULT_MAIL_WEEKLY_DAY = 5;                    // Friday
 var DEFAULT_MAIL_WEEKLY_HOUR = 12;                  // Noon
-var MINIMUM_MAIL_POLLING_FREQUENCY = 60;            // 1 minute
+var MINIMUM_MAIL_POLLING_FREQUENCY = 1;            // 1 minute
```
nicolaasmatthijs: Alignment

node_modules/oae-tenants/lib/api.js
```diff
@@ -456,6 +457,7 @@ var _updateCachedTenant = function(tenantAlias, callback) {
  * @param  {String}     host                    The host on which this tenant will be proxying (e.g. oae.cam.ac.uk or oae.gatech.edu)
  * @param  {Object}     [opts]                  Optional arguments
  * @param  {String}     [opts.emailDomain]      The email domain for this tenant
+ * @param  {String}     [opts.countryCode]      The ISO-3166-1 country code of the country that represents the tenant
```
nicolaasmatthijs: We document `ISO-3166-1` but are requiring `iso-3166-2`. Is there any significant difference ?

node_modules/oae-tenants/lib/rest.js
```diff
@@ -206,13 +206,7 @@ OAE.globalAdminRouter.on('post', '/api/tenantNetwork/:id/removeTenants', functio
...
  */
 OAE.globalAdminRouter.on('post', '/api/tenant/create', function(req, res) {
-    var opts = {};
-
-    // Apply the email domain if it is specified
-    if (_.isString(req.body.emailDomain)) {
-        opts.emailDomain = req.body.emailDomain;
-    }
-
+    var opts = _.oaeExtendDefined({}, _.pick(req.body, 'emailDomain', 'countryCode'));
```
nicolaasmatthijs: I think the API already deals with unexpected values being passed in

node_modules/oae-tenants/lib/api.js
mrvisser: The `iso-3166-2` module supports `ISO-3166-1 alpha 2` country codes :/

node_modules/oae-tenants/lib/api.js
```diff
@@ -14,6 +14,7 @@
  */
 
 var _ = require('underscore');
+var iso3166 = require('iso-3166-2');
```
nicolaasmatthijs: It doesn't look like we're actually using this. We're only using the validator wrapper in this file

nicolaasmatthijs:
@mrvisser : Can you also create a UI counterpart that adds the country to the tenant list and tenant creation?

nicolaasmatthijs:
Finished review. Back to @mrvisser for follow-up.

node_modules/oae-tenants/lib/rest.js
mrvisser: The API will throw an error if misused, however `alias` and `displayName`, for example, are not valid, so we have to pick these off.

mrvisser:
Followed up, back to @nicolaasmatthijs 

UI portion is in progress

nicolaasmatthijs:
Looks good. Merged.

PR #1232, opened 2015-09-19
(.gitignore) Add workaround for atom fuzzy-finder index: https://github.com/atom/fuzzy-finder/issues/65

mrvisser:


mrvisser:

.gitignore
```diff
@@ -1,9 +1,11 @@
 .DS_STORE
 dump.rdb
 target/
+*.log
+doc_data.txt
+
 node_modules/*
 !node_modules/oae-*
-node_modules/oae-rest
+!node_modules/oae-*/*
```
simong: Why is this required? Wouldn't `!node_modules/oae-*` cover this?

mrvisser: For git, yes. But this is the atom bug workaround.

simong: Can you add a comment as it not immediately obvious and an oaf like me might remove it when trying to micro optimize?

mrvisser:
I accidentally pushed this fix to master with a README.md change. I've followed up on this here: https://github.com/oaeproject/Hilary/commit/9711b11eb93583a9b73a03be8fb326a937dd5463

PR #1233, opened 2015-09-21
Deleted user account is not being scrubbed in members list

stuartf:
Currently if you create a user, add them to a group, then delete them, their entry should remain in the group members list, but have its profile path and display name "scrubbed" / obfuscated.

It appears that is currently not the case.

### DIFF
--- a/node_modules/oae-principals/lib/util.js
+++ b/node_modules/oae-principals/lib/util.js
@@ -184,15 +184,15 @@ var hideUserData = module.exports.hideUserData = function(ctx, user) {
     var isLoggedIn = TenantsUtil.isLoggedIn(ctx, user.tenant.alias);
     var isTargetUser = (!isAnon && ctx.user().id === user.id);
     var isAdmin = (!isAnon && ctx.user().isAdmin && ctx.user().isAdmin(user.tenant.alias));
+    var needsLoggedIn = user.visibility === AuthzConstants.visibility.LOGGEDIN;
+    var isPrivate = user.visibility === AuthzConstants.visibility.PRIVATE;
 
     if (isAdmin || isTargetUser) {
         return user;
     }
 
     // Hide the sensitive profile information if the user has limited access
-    if ((user.deleted ||
-        user.visibility === AuthzConstants.visibility.LOGGEDIN && !isLoggedIn) ||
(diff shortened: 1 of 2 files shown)


node_modules/oae-principals/lib/util.js
```diff
@@ -190,7 +190,8 @@ var hideUserData = module.exports.hideUserData = function(ctx, user) {
     }
 
     // Hide the sensitive profile information if the user has limited access
-    if ((user.visibility === AuthzConstants.visibility.LOGGEDIN && !isLoggedIn) ||
+    if ((user.deleted ||
+        user.visibility === AuthzConstants.visibility.LOGGEDIN && !isLoggedIn) ||
         (user.visibility === AuthzConstants.visibility.PRIVATE && !isTargetUser)) {
```
mrvisser: Can you please wrap some of these booleans into `isSomething` variables to make the if statement a bit more understandable?

mrvisser:
This needs unit tests

mrvisser:
Back to @stuartf for follow up

stuartf:
Ready for review

node_modules/oae-principals/tests/test-delete.js
```diff
@@ -657,36 +657,52 @@ describe('Principals Delete and Restore', function() {
...
                 assert.ok(!err);
                 TestsUtil.generateTestGroups(camAdminRestContext, 1, function(group) {
 
-                    // Add the user to the group members library
-                    var roleChanges = {};
-                    roleChanges[user.user.id] = 'member';
-                    PrincipalsTestUtil.assertSetGroupMembersSucceeds(camAdminRestContext, camAdminRestContext, group.group.id, roleChanges, function() {
+                    // Change user1's publicAlias
+                    RestAPI.User.updateUser(user1.restContext, user1.user.id, {'publicAlias': 'Clark Kent'}, function(err, user) {
+                        console.log(err);
```
sallakarppinen: Should be removed.

sallakarppinen:
Currently the deleted user's profile path is still visible when viewing group members as an administrator; as the profile path for a deleted user will always 404, I think it should be scrubbed even for admins.

brecke:
> Currently the deleted user's profile path is still visible when viewing group members as an administrator; as the profile path for a deleted user will always 404, I think it should be scrubbed even for admins.

Added to #1304

PR #1236, opened 2015-09-23
(InvitingGuests) Fix issue where one resource is removed from email invitation

mrvisser:


mrvisser:

PR #1238, opened 2015-09-25
Always given a validated email when accepting an invitation

mrvisser:


mrvisser:
Assigning to @simong for review. @nicolaasmatthijs feel free to pick it up instead if you get time tonight.

node_modules/oae-authentication/lib/strategies/facebook/init.js
```diff
@@ -65,12 +66,7 @@ module.exports = function() {
                 opts.email = profile.emails[0].value;
             }
 
-            if (opts.email) {
-                opts.emailVerified = true;
-            }
-
-            var context = new Context(tenant, null);
```
simong: We can get rid of the `var Context = require('oae-context');` line here and in a few similar files

node_modules/oae-authentication/lib/util.js
```diff
@@ -227,6 +230,40 @@ var validateRedirectUrl = module.exports.validateRedirectUrl = function(redirect
...
+ */
+var handleExternalGetOrCreateUser = module.exports.handleExternalGetOrCreateUser = function(req, authProvider, externalId, providerProperties, displayName, opts, callback) {
+    if (!opts.email) {
+        // If no email was provided by the external authentication provider,
+        // we should provide the invitation token to indicate that the user
+        // should be created with their email address pre-validated to that
+        // associated to the invitation token
+        var invitationInfo = _getRequestInvitationInfo(req);
+        if (invitationInfo.invitationToken) {
+            opts.invitationToken = invitationInfo.invitationToken;
```
simong: Can you add a test case that tests this logic? There's `AuthenticationTestUtil.assertGoogleLoginSucceeds` that should help with this

simong:
LGTM, couple of small things, back to @mrvisser

mrvisser:
Added a test for an external provider not providing an email address when there is and isn't an invitation token.

Google can't be used as an external auth provider that doesn't return an email because the email is the external id so it is required. I didn't use Twitter because it uses OAuth 1.1 which I'm assuming would require signing responses. Facebook seemed like the best choice.

mrvisser:
back to @simong for review

PR #1239, opened 2015-09-28
Create script to identify accounts with duplicate emails and their activity

mrvisser:


nicolaasmatthijs:
@mrvisser : Is this ready for review?

nicolaasmatthijs:

etc/scripts/analysis/user-activity.js
```diff
@@ -0,0 +1,510 @@
...
+    /*!
+     * Given a paged result set of principal hashes, filter them down to those with email addresses
+     * and those that are part of the specified tenant. Then add them to the shared `userHashes`
+     * array
+     *
+     * @param  {Object[]}   principalHashes     The principals to filter and aggregate
+     * @param  {Function}   callback            Will be invoked when the principals are aggregated
+     */
+    function _aggregateUsers(principalHashes, callback) {
+        log().info('Analyzing %s principals to index by email', principalHashes.length);
```
simong: Wouldn't this just print `Analyzing 100 principals to index by email` for every batch? Seems a bit redundant if we have one a bit lower in the `.tap` statement that reports actual numbers?

etc/scripts/analysis/user-activity.js
```diff
@@ -0,0 +1,510 @@
...
+
+    /*!
+     * Given a paged result set of principal hashes, filter them down to those with email addresses
+     * and those that are part of the specified tenant. Then add them to the shared `userHashes`
+     * array
+     *
+     * @param  {Object[]}   principalHashes     The principals to filter and aggregate
+     * @param  {Function}   callback            Will be invoked when the principals are aggregated
+     */
+    function _aggregateUsers(principalHashes, callback) {
```
simong: Even though this is a deviation from the typical OAE styleguide, I'm quite in favor of declaring private/inner functions like this.

Any objections @stuartf / @nicolaasmatthijs  ?

etc/scripts/analysis/user-activity.js
```diff
@@ -0,0 +1,510 @@
...
+            .tap(function(userHashes) {
+                log().info('Preparing to write a total of %s users who have duplicate emails', userHashes.length);
+            })
+
+            // Re-group things into sane concurrent batches
+            .groupBy(function(userHash, i) { return Math.floor(i / 5); })
+            .values()
+            .value();
+
+        log().info('Begin writing %s batches of user activity to CSV', batches.length);
```
simong: It might be useful to report the number of users that go into a batch, as there will always be 5 batches?

simong: Oops, misread `/` as `%` and was getting worries that we'd end up with batches of ~5000 users for a minute. Ignore the above!

etc/scripts/analysis/user-activity.js
```diff
@@ -0,0 +1,510 @@
...
+    var userHashBatch = userHashBatches.shift();
+    var _done = _.after(userHashBatch.length, function() {
+        log().info('Successfully completed batch of %d users', userHashBatch.length);
+        return _writeUserRows(userHashBatches);
+    });
+
+    _.each(userHashBatch, function(userHash) {
+        _createUserRow(userHash, function(err, row) {
+            if (err) {
+                log().warn({'err': err, 'user': userHash}, 'Failed to create row from a user hash');
```
simong: Should we add a blank entry to the CSV file for that user?

simong:
LGTM, if staging is still around, we could give this a quick whirl on there first?

mrvisser:
> LGTM, if staging is still around, we could give this a quick whirl on there first?

Staging has no data so it's not a great test case. I'll run it on QA first. But I have run it locally, and I ran it on a full staging snapshot months ago.

etc/scripts/analysis/user-activity.js
mrvisser: I think I prefer the explicitness given this is a script someone will run and watch. If someone starts running the script and they potentially just see:

`Indexing 0 users with emails for the specified tenant (if any)`

... a bunch for the first little while, it might look like something is wrong. If we indicate that we're scanning non-0 batches of principals, then they have a better idea that there is something happening.

etc/scripts/analysis/user-activity.js
stuartf: It seems fine to me. AFAIK the only practical difference will be how the function gets hoisted, but I think that actually works out better this way.

nicolaasmatthijs: Is there a reason to deviate from the style guide other than stylistic preference? I don't have a strong opinion on this and quite like the idea of using this style (especially for inner functions), but it might be easier to just follow the existing guidelines here.

mrvisser: > Is there a reason to deviate from the style guide other than stylistic preference?

The style guide is built on the assumption of using module-based invocation, where there is a run of the code first before any actual method invocation is performed (e.g., when you do your first `require`, all you're normally doing is declaring your variables / functions in the file and exporting functionality, not usually invoking actual functionality). Because of this, we can define public functions at the top, and private functions at the bottom, which I believe is part of our stylistic guideline (and a more significant guideline than exactly how functions are declared).

In these scripts, the environment is different as the first and only invocation is the one that is invoking functions. We have to either make a decision to violate our style guidelines by declaring all the very trivial internal functionality at the top (e.g., a function to split a string in a certain way), or violate it by just declaring the functions using `function`, and I chose to declare the functions differently because every time I open up a script I don't want to have to scroll down to the bottom to find the substance of the script functionality.

So yes it was a choice based on stylistic preference, but we were going to have to violate "style guidelines" some how.

nicolaasmatthijs: That makes sense. I'm fine with it too.

PR #1240, opened 2015-09-28
Emails still being sent to users that were deleted

mrvisser:
It appears the issue fixed in #1216 is no longer taking effect.

At the least, this change seems to have been reverted while pulling inviting-guests into master: https://github.com/oaeproject/Hilary/pull/1216/files#diff-603d08a65509e4f3d08aae9970eaafa0R94

There is obviously also an issue w/ the unit tests that was created for it since it is not failing as a result of this.

### DIFF
--- a/node_modules/oae-activity/tests/test-email.js
+++ b/node_modules/oae-activity/tests/test-email.js
@@ -21,7 +21,8 @@ var util = require('util');
 
 var AuthzUtil = require('oae-authz/lib/util');
 var ConfigTestUtil = require('oae-config/lib/test/util');
-var EmailTestUtil = require('oae-email/lib/test/util');
+var ContentTestUtil = require('oae-content/lib/test/util');
+var EmailTestUtil= require('oae-email/lib/test/util');
 var MqTestUtil = require('oae-util/lib/test/mq-util');
 var PrincipalsTestUtil = require('oae-principals/lib/test/util');
 var RestAPI = require('oae-rest');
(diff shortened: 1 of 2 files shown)


mrvisser:

mrvisser:
> There is obviously also an issue w/ the unit tests that was created for it since it is not failing as a result of this.

The reason the unit test wasn't catching this was because there was only 1 user queued to have email sent to them. When that's the case, PrincipalsDAO returns the full user profile, regardless of the fields being specified. So this bug only manifests when there are multiple users to send email to, which the unit test previously did not cover.

An issue has been opened to make the result of PrincipalsDAO.getPrincipals consistent regardless of the number of principal ids specified.

node_modules/oae-activity/tests/test-email.js
```diff
@@ -860,46 +861,52 @@ describe('Activity Email', function() {
...
-            // Set the appropriate email preference for mrvisser
-            RestAPI.User.updateUser(mrvisser.restContext, mrvisser.user.id, {'emailPreference': 'immediate'}, function(err) {
-                assert.ok(!err);
+                // Ensure we have no emails queued to send
+                EmailTestUtil.collectAndFetchAllEmails(function() {
 
-                // Generate an email activity for mrvisser
-                RestAPI.Content.createLink(simong.restContext, 'Google First', 'Awesome Google', 'public', 'http://www.google.ca/firstlink', [mrvisser.user.id], [], [], function(err, link) {
-                    assert.ok(!err);
+                    // Generate an email activity for the group, ensuring both
```
simong: `, ensuring both users receive an email`?

node_modules/oae-activity/tests/test-email.js
```diff
@@ -860,46 +861,52 @@ describe('Activity Email', function() {
...
+                            // Ensure the 2 recipients are mrvisser and nico
+                            var recipients = _.chain(messages)
+                                .pluck('headers')
+                                .pluck('to')
+                                .value()
+                                .sort();
+                            assert.deepEqual(recipients, expectedRecipients);
 
-                            // Delete mrvisser now with pending emails in their notifications feed
+                            // Delete and ensure the same action results in only nico receiving the email
```
simong: `Delete mrvisser and ensure the same ...`

simong:
Looks OK to me, feel free to merge into master when/if you fix up those comments

PR #1241, opened 2015-10-03
Only allow local account creation with configured email domain

simong:
When a tenant is configured with an email domain, it should only allow local account creation with that configured email domain.


simong:
Attached a commit that does this

nicolaasmatthijs:

node_modules/oae-authentication/lib/api.js
```diff
@@ -331,6 +339,30 @@ var _getOrCreateUser = function(ctx, loginId, displayName, opts, callback) {
...
 /**
+ * Validate that an email address belongs to the tenant in context
+ *
+ * @param  {Context}    ctx             Standard context object containing the current user and the current tenant
+ * @param  {String}     email           The email address to validate
+ * @param  {Function}   callback        Standard callback function
+ * @param  {Object}     callback.err    An error that occurred, if any
+ * @api private
+ */
+var validateEmailBelongsToTenant = function(ctx, email, callback) {
```
mrvisser: needs prefix with `_`

node_modules/oae-authentication/lib/api.js
```diff
@@ -307,12 +308,19 @@ var _getOrCreateUser = function(ctx, loginId, displayName, opts, callback) {
                     opts.emailVerified = true;
                 }
 
-                createUser(ctx, loginId, displayName, opts, function(err, user) {
+                var shouldCheckEmail = (opts.email && !opts.authoritative);
+                OaeUtil.invokeIfNecessary(shouldCheckEmail, validateEmailBelongsToTenant, ctx, opts.email, function(err) {
```
mrvisser: I think if a non-authoritative authentication method doesn't provide an email (e.g., Twitter), we should still restrict login if the email domain of the tenant is set. Otherwise, this would mean that the entire public can come in VIA twitter, which we're trying to avoid.

node_modules/oae-principals/lib/api.user.js
```diff
@@ -826,6 +826,9 @@ var getMe = module.exports.getMe = function(ctx, callback) {
     // Indicate whether the tenant is private or not
     tenant.isPrivate = TenantsUtil.isPrivate(tenant.alias);
 
+    // Add the email domain to the tenant object
+    tenant.emailDomain = ctx.tenant().emailDomain;
```
mrvisser: Should `tenant.compact()` be changed to not remove this?

node_modules/oae-principals/tests/test-users.js
```diff
@@ -326,13 +327,74 @@ describe('Users', function() {
...
+            var anonymousRestContext = TestsUtil.createTenantRestContext(tenantHost);
+            RestAPI.User.createUser(anonymousRestContext, username, 'password', 'Test User', email, {'visibility': 'public'}, function(err) {
+                if (expectSuccess) {
+                    assert.ok(!err);
+                } else {
+                    assert.ok(err);
+                }
+                return callback();
+            });
+        };
```
mrvisser: Can you extract `expectSuccess` into an `opts` object instead, so it's not just a cryptic boolean in the parameter list?

Also, it's possible we can just use a `assertCreateUserSucceeds` / `assertCreateUserFails` instead of defining a new function.

mrvisser:
Finished review, back to @simong

node_modules/oae-authentication/lib/api.js
simong: What should we do in that case? Block the user from creating an account?

simong:
I've pushed some follow-up except for the "no email provided when signing in with Twitter if there's an email domain configured". I'm not entirely sure what the expected response should be

node_modules/oae-authentication/lib/api.js
mrvisser: Any reason not to do the same thing we would do if the non-authoritative external login method provided an email that didn't match the tenant's email domain ?

node_modules/oae-authentication/lib/api.js
```diff
@@ -307,12 +308,31 @@ var _getOrCreateUser = function(ctx, loginId, displayName, opts, callback) {
...
+
+                // However, if a user followed a link from an invitation email we do not check whether
+                // the email belongs to the configured tenant's email domain. This is to allow for the
+                // situation where a tenant's email domain gets changed after a user gets invited
+                // but before the user has accepted the invitation
+                if (opts.invitationToken && opts.emailVerified) {
+                    shouldCheckEmail = false;
+                }
+
+                OaeUtil.invokeIfNecessary(shouldCheckEmail, _validateEmailBelongsToTenant, ctx, opts.email, function(err) {
```
nicolaasmatthijs: From @mrvisser : "Any reason not to do the same thing we would do if the non-authoritative external login method provided an email that didn't match the tenant's email domain ?"

nicolaasmatthijs: How does signing in using a non-authoritative external login method that doesn't provide an email that matches the tenant's email domain currently manifest itself in the UI?

simong: The same way we've always done it. We create an account without an email address and then pester the user until he fills one in. I suppose we could expand the scope of this PR and block the user from signing in...

mrvisser: I don't understand how that is expanding the scope of this PR. If someone logs in with GMail, doesn't this PR block their signin too with the `validateEmailBelongsToTenant` function?

simong:
I pushed some follow-up that will effectively block the account from being created when:
 - the tenant has a configured email domain
 - and the user signed in through anything other than CAS or Shibboleth
 - and no email address was released or an email address was released that didn't match the configured tenant's email domain

The UI does not deal with this just yet

mrvisser:
Sorry, accidentally commented on the commit. There are a couple small things remaining then this is good to go. Back to @simong

simong:
Pushed more follow-up

mrvisser:
Looks good!

PR #1244, opened 2015-10-05
Use a CPP Cassandra driver

simong:


config.js
```diff
@@ -33,7 +33,7 @@ config.ui = {
 
 // Cassandra related config information.
 config.cassandra = {
-    'hosts': ['127.0.0.1:9160'],
+    'hosts': '127.0.0.1',
```
mrvisser: :confetti_ball:

node_modules/oae-activity/lib/internal/dao.js
```diff
@@ -77,23 +77,26 @@ var getActivities = module.exports.getActivities = function(activityStreamId, st
...
+    });
+    var query = util.format('SELECT * FROM "ActivityStreams" WHERE "activityStreamId" IN (%s)', placeholders.join(','));
+    var parameters = activityStreamIds.slice();
     if (start) {
         query += ' AND "activityId" > ?';
         parameters.push(start + ':');
     }
 
-    Cassandra.runQuery(query, parameters, function(err, rows) {
+    Cassandra.runQuery(query, parameters, {'cache': false}, function(err, rows) {
```
mrvisser: Is it possible that, if we use `runInQuery` query for this kind of case and never cache IN queries, that we would not have to add an additional parameter to `runQuery`?

node_modules/oae-principals/lib/internal/dao.js
```diff
@@ -81,9 +81,9 @@ var createUser = module.exports.createUser = function(user, callback) {
  * @param  {Group}      callback.group      The created group
  */
 var createGroup = module.exports.createGroup = function(groupId, tenantAlias, displayName, description, visibility, joinable, createdBy, callback) {
-    var lastModified = Date.now();
+    var lastModified = new Date();
     var query = 'INSERT INTO "Principals" ("principalId", "tenantAlias", "displayName", "description", "visibility", "joinable", "lastModified", "createdBy", "created") VALUES (?, ?, ?, ?, ?, ?, ?, ?, ?)';
-    var parameters = [groupId, tenantAlias, displayName, description, visibility, joinable, lastModified.toString(), createdBy, lastModified.toString()];
+    var parameters = [groupId, tenantAlias, displayName, description, visibility, joinable, '' + lastModified.getTime(), createdBy, lastModified];
```
mrvisser: Is this equivalent to `lastModified.getTime().toString()`? I'd prefer that if possible

node_modules/oae-tests/lib/util.js
```diff
@@ -876,8 +876,9 @@ var _createGroupWithVisibility = function(tenant, visibility, memberPrincipalId,
     var randomId = util.format('%s-%s', visibility, ShortId.generate());
     var displayName = 'displayName-' + randomId;
     var description = 'description-' + randomId;
+
     RestAPI.Group.createGroup(tenant.adminRestContext, displayName, description, visibility, 'request', [], [memberPrincipalId], function(err, newGroup) {
-        assert.ok(!err);
+        assert.ok(!err, JSON.stringify(err));
```
mrvisser: Oop ;)

node_modules/oae-util/lib/cassandra.js
```diff
@@ -347,14 +271,53 @@ var _createColumnFamilies = function(keys, families, callback) {
...
  * @param  {Object}   callback.err  An error that occurred, if any
  */
-var runQuery = module.exports.runQuery = function(query, parameters, callback) {
+var runQuery = module.exports.runQuery = function(query, parameters, opts, callback) {
+    if (typeof opts === 'function') {
+        callback = opts;
+        opts = {
+            'cache': true
+        };
+    }
```
mrvisser: I see what's being done here. I'm ok w/ it I think. Docs do need to be updated, and I wonder if there is a good way to highlight this style of function in the docs.

If we can, however, get rid of this parameter entirely with my previous comment, I would be in favour of that instead.

node_modules/oae-util/lib/cassandra.js
```diff
@@ -742,17 +697,58 @@ var constructUpsertCQL = module.exports.constructUpsertCQL = function(cf, rowKey
...
+            //    avoid leaking memory when someone builds a dynamic query but forgets to set the
+            //    `opts.cache` value to false
+            if (opts.cache && query.split('?').length < 10) {
+                cachedPreparedQueries[query] = preparedQuery;
+            }
+
+            return callback(null, preparedQuery);
+        });
+    }
+};
```
mrvisser: How much performance benefit is prepared statements giving us? Do you think it's really necessary, in favour of maybe a more explicit way of registering prepared statements w/ the driver?

E.g., sometime in the future, exposing a function `Cassandra.prepare` for a statement, and then high-frequency queries such as authz and activity writes can just prepare their statements on init, and then there wouldn't be any other code or heuristics associated to it.

simong:
Note to self:
> Add a migration unit test setup similar to what @mrvisser  did for the CQL2 -> CQL3 migration?
> A script that will seed the test keyspace with a production dump of the schema, then do the migration, then run the unit tests overtop of it

PR #1245, opened 2015-10-06
Allow for templatization of the email From header

simong:
Allow for the email From header configuration to contain `${tenant}`, which is then replaced by the tenant display name. This will be used to achieve the new default From header: `*Unity for <tenantName>` without requiring a custom configuration value for every tenant.


simong:
Attached a commit

nicolaasmatthijs:

mrvisser:
Looks good.

PR #1246, opened 2015-10-07
(WAYF) Add isGuestTenant to tenant model

mrvisser:


node_modules/oae-principals/lib/test/util.js
```diff
@@ -270,6 +270,18 @@ var assertGetUserFails = module.exports.assertGetUserFails = function(restCtx, u
 var assertGetMeSucceeds = module.exports.assertGetMeSucceeds = function(restCtx, callback) {
     RestAPI.User.getMe(restCtx, function(err, me, response) {
         assert.ok(!err);
+
+        // Sanity check the me model
+        assert.ok(me);
+        assert.ok(me.tenant);
+        assert.ok(_.isString(me.tenant.alias));
+        assert.ok(_.isString(me.tenant.displayName));
+        if (me.tenant.alias === 'guest') {
```
nicolaasmatthijs: Should we get this value from the config file ?

node_modules/oae-principals/lib/test/util.js
```diff
@@ -270,6 +270,18 @@ var assertGetUserFails = module.exports.assertGetUserFails = function(restCtx, u
...
+            assert.strictEqual(me.tenant.isGuestTenant, true);
+        } else {
+            assert.ok(!_.chain(me.tenant).keys().contains('isGuestTenant').value());
```
nicolaasmatthijs: Can't we just do `assert.ok(!me.tenant.isGuestTenant)` / `assert.ok(_.isUndefined(me.tenant.isGuestTenant))` ?

node_modules/oae-tenants/lib/model.js
```diff
@@ -24,6 +24,7 @@
  * @param  {String}     [opts.emailDomain]          The tenant's registered email domain
  * @param  {String}     [opts.countryCode]          The ISO-3166-1 country code of the country that represents the tenant
  * @param  {Boolean}    [opts.isGlobalAdminServer]  Whether or not the tenant is the global admin tenant. Default: `false`
+ * @param  {Boolean}    [opts.isGuestTenant]        Whether or nott he tenant is the tenant configured as the "guest" tenant. Default: `false`
```
nicolaasmatthijs: typo

node_modules/oae-tenants/lib/model.js
```diff
@@ -37,6 +38,7 @@ var Tenant = module.exports.Tenant = function(alias, displayName, host, opts) {
     that.countryCode = opts.countryCode;
     that.active = (opts.active !== false);
     that.isGlobalAdminServer = (opts.isGlobalAdminServer === true);
```
nicolaasmatthijs: We probably don't need to pass in `opts.isGlobalAdminServer` and `opts.isGuestTenant` here, as we can just infer that from the config file.

node_modules/oae-tenants/tests/test-tenants.js
```diff
@@ -352,6 +354,24 @@ describe('Tenants', function() {
                 });
             });
         });
+
+        it('verify guest tenant is flagged in the me feed', function(callback) {
```
nicolaasmatthijs: This needs some JSDocs to be in line with the other tests

node_modules/oae-tenants/tests/test-tenants.js
```diff
@@ -352,6 +354,24 @@ describe('Tenants', function() {
...
+            var guestRestContext = TestsUtil.createTenantRestContext('guest.oae.com');
+            PrincipalsTestUtil.assertGetMeSucceeds(guestRestContext, function(me) {
+                assert.strictEqual(me.tenant.isGuestTenant, true);
+
+                // Ensure the isGuestTenant key does not exist when it is not
+                // the guest tenant
+                TestsUtil.generateTestUsers(camAdminRestContext, 1, function(err, users, user) {
+                    assert.ok(!err);
+                    PrincipalsTestUtil.assertGetMeSucceeds(user.restContext, function(me) {
+                        assert.ok(!_.chain(me.tenant).keys().contains('isGuestTenant').value());
```
nicolaasmatthijs: Same here as above: `assert.ok(!me.tenant.isGuestTenant)` / `assert.ok(_.isUndefined(me.tenant.isGuestTenant))` ?

nicolaasmatthijs:
Finished review. Back to @mrvisser for follow-up

node_modules/oae-principals/lib/test/util.js
mrvisser: I'd rather more strictly test that the key doesn't somehow get added to the response with a falsey value.

node_modules/oae-tenants/lib/model.js
mrvisser: We could, but the model object doesn't have access to the config, nor would we really want to require that the `Tenant` model be "initialized" with a copy of the config on startup.

node_modules/oae-principals/lib/test/util.js
nicolaasmatthijs: That's what the `assert.ok(_.isUndefined(me.tenant.isGuestTenant))` one does right? It verifies the key isn't present.

mrvisser: It ensures that the value isn't `undefined`, but in JS it's possible to have a key whose value is set to `undefined`.

Though I didn't consider though that it's coming from a JSON response, so it's probably fine to do `_.isUndefined`.

node_modules/oae-principals/lib/test/util.js
mrvisser: Ideally yes, but we currently don't expose this information anywhere and we hardcode `'admin'` in all other areas where the global admin tenant alias is needed.

mrvisser:
This is now ready for review. To @nicolaasmatthijs

nicolaasmatthijs:
Looks good. Merged

PR #1248, opened 2015-10-16
Make tenant email domain lower-cased when stored

mrvisser:
Currently it validates and does everything case-insensitive. But the stored version is original case. Perhaps it should be lower cased in storage.

### DIFF
--- a/node_modules/oae-authz/tests/test-invitations.js
+++ b/node_modules/oae-authz/tests/test-invitations.js
@@ -680,6 +680,41 @@ describe('Invitations', function() {
         });
 
         /**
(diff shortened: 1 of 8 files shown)


mrvisser:
There was a regression in inviting guests that resulted in host and email domain not being lower-cased when updated. For host this was fine because there was secondary lower-casing when reading the tenant from the DB.

This fix:

* Addresses the issue where the host and email domain aren't properly lower cased into the db
* Adds the same secondary lower-casing on read for email domain that we already have for host
* Updates the email domain index to work case-insensitive, which is proper for domain indexing since domains are by nature case insensitive

PR #1249, opened 2015-10-20
Template the logo URL in the skin

mrvisser:


mrvisser:

node_modules/oae-ui/lib/api.js
```diff
@@ -703,7 +703,37 @@ var getTenantSkinVariables = module.exports.getTenantSkinVariables = function(te
...
+ * @param  {Object}     obj     The arbitrary object whose value to render
+ * @param  {String}     key     The key of the specified object to replaced with a rendered version
+ * @param  {Object}     data    The data for the template
+ * @api private
+ */
+var _renderDynamicValue = function(obj, key, data) {
+    var value = obj[key];
+    if (value) {
+        try {
+            obj[key] = _.template(value)(data);
```
simong: In the past we haven't exposed underscore templates to tenant admins as they could enter stuff like:

```javascript
var skinConfig = {
    'institutional-logo-url': '   '<% while (true) { console.log('LOL'); } %>'
};
```

and effectively block the server

mrvisser: Of course they can, great catch. What was I thinking?

simong:
Other than the underscore templates, this looks good. FWIW, I think the CAS authentication does a similar thing to construct a display name. Maybe use the same simple format?

Assigning back to @mrvisser

mrvisser:
Followed up. I made it the same as the from header template that relies on `${tenant}`. But perhaps in a later version we should converge on `{key}`

PR #1250, opened 2015-10-21
There is an unused folder visibility configuration in the content module

mrvisser:
There appears to be a configuration for folders visibility in both the `oae-content/config/content.js` and `oae-folders/config/folders.js` file.

Only the `oae-folders/config/folders.js` is used, so we should be able to just remove the oae-content version of this configuration.


mrvisser:

PR #1264, opened 2015-11-12
(Issue 1260 / Aggregation Loop) Ensure a broken activity does not result in an infinite aggregation loop

mrvisser:
This is best reviewed with `?w=1`

### DIFF
--- a/node_modules/oae-activity/lib/internal/dao.js
+++ b/node_modules/oae-activity/lib/internal/dao.js
@@ -240,11 +240,12 @@ var saveQueuedActivities = module.exports.saveQueuedActivities = function(activi
 /**
  * Get the queued activities from the given bucket number for processing.
  *
- * @param  {Number}    bucketNumber                The bucket from which to fetch the queued activities
- * @param  {Number}    limit                       How many routed activities to fetch from the queue
- * @param  {Function}  callback                    Standard callback function
- * @param  {Object}    callback.err                An error that occurred, if any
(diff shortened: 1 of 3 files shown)


mrvisser:

node_modules/oae-activity/lib/internal/dao.js
```diff
@@ -256,23 +256,29 @@ var getQueuedActivities = module.exports.getQueuedActivities = function(bucketNu
...
-            queuedActivities[_createRoutedActivityKey(routedActivity)] = routedActivity;
+                log().error({'err': err, 'routedActivity': truncatedRoutedActivity}, 'Error trying to parse stored routed activity');
+            }
         });
 
-        log().trace({'queuedActivities': queuedActivities}, 'Fetched queued activities.');
+        log().trace({'queuedActivities': queuedActivities}, 'Fetched queued activities');
 
-        return callback(null, queuedActivities);
+        return callback(null, queuedActivities, numToDelete);
```
simong: It probably doesn't matter for the aggregator but if one of the activities couldn't be parsed this `numToDelete` value will be off by one. Additionally, can you add it in the jsdoc?

simong:
LGTM, back to @mrvisser for small fix

node_modules/oae-activity/lib/internal/dao.js
mrvisser: That's intentional, because we want to delete items that can't be parsed, so they should be included in the "num to delete" count.

mrvisser:
Followed up, back to @simong

mrvisser:
Going to go ahead and merge in time for the bug bash.

PR #1265, opened 2015-11-13
Add Mixpanel instrumentation for invites

simong:
We should have a Mixpanel invite event that tracks at least:

- The inviter id
- The invite token
- The tenant from which the invite is sent
- The tenant to which the user will be invited
- The email domain that's been used for the invite
- The action type for which the invite was sent (group, content, etc.)

We should also have a Mixpanel event for accepting an invitation:

- The invite token
- The tenant to which the user was invited
- The tenant at which the invitation was accepted

### DIFF
--- a/node_modules/oae-mixpanel/lib/eventlisteners/invitations.js
+++ b/node_modules/oae-mixpanel/lib/eventlisteners/invitations.js
@@ -1,0 +1,85 @@
+/*!
+ * Copyright 2015 Apereo Foundation (AF) Licensed under the
+ * Educational Community License, Version 2.0 (the "License"); you may
+ * not use this file except in compliance with the License. You may
+ * obtain a copy of the License at
+ *
+ *     http://opensource.org/licenses/ECL-2.0
+ *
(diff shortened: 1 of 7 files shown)


node_modules/oae-mixpanel/lib/eventlisteners/invitations.js
```diff
@@ -0,0 +1,85 @@
...
+    /**
+     * One or more invitations got sent out
+     */
+    MixpanelUtil.listen(ResourceActions, ResourceConstants.events.INVITED, function(ctx, invitations, emailTokens) {
+        _.each(invitations, function(invitation) {
+            var params = MixpanelUtil.getBasicParameters(ctx);
+            params.inviteToken = emailTokens[invitation.email];
+
+            // The user who is inviting the guest
+            params.invitedBy = ctx.user().id;
```
nicolaasmatthijs: `invitedBy` is already part of the basic mixpanel parameters, so no need to repeat it here

PR #1255, opened 2015-11-16
Explicitly set session cookie name

simong:
This change:

https://github.com/expressjs/cookie-session/commit/4c55ae4fb68377428c378f54515032220eed81e1#diff-168726dbe96b3ce427e7fedce31bb0bcL44

Has resulted in the session cookie to change name to `session` instead of `expess.sess`. We should fix this name to `session` in case it changes again in another version.

### DIFF
--- a/node_modules/oae-util/tests/test-server.js
+++ b/node_modules/oae-util/tests/test-server.js
@@ -178,8 +178,12 @@ describe('OAE Server', function() {
                     var config = require('../../../config.js').config;
                     assert.ok(_.contains(cookieNames, config.cookie.name));
 
-                    // Remove the session cookie from the jar
-                    delete simong.restContext.cookieJar._jar.store.idx.localhost['/'][config.cookie.name];
+                    // Rename the cookie to something else
+                    _.each(cookies, function(cookie) {
+                        if (cookie.key === config.cookie.name) {
+                            cookie.key = 'somethingElse';
+                        }
+                    });
 
                     // Ensure the user is anonymous
                     PrincipalsTestUtil.assertGetMeSucceeds(simong.restContext, function(me) {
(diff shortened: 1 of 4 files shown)


mrvisser:
Can you have the tests use a cookie name that is different than the express default, and ensure it is the name of the cookie in the tests?

simong:
Requires https://github.com/oaeproject/oae-rest/pull/27

node_modules/oae-util/tests/test-server.js
```diff
@@ -155,4 +156,43 @@ describe('OAE Server', function() {
...
+                    // Verify the configured cookie names are being used
+                    var config = require('../../../config.js').config;
+                    assert.ok(_.contains(cookieNames, config.cookie.name));
+
+                    // Rename the cookie to something else
+                    _.each(cookies, function(cookie) {
+                        if (cookie.key === config.cookie.name) {
+                            cookie.key = 'somethingElse';
+                        }
+                    });
```
mrvisser: Rather than changing the cookie on the client, I was referring to changing the `config.cookie.name` value in the initial test setup to something like `somethingElse`. I want to avoid the case where express has another major version change in which our `config.cookie.name` is no longer taking effect and our tests don't catch it because we're using the default.

mrvisser:
There is a miscommunication about the test setup, back to @simong

simong:
Pushed some follow-up

mrvisser:
```
$ grunt
Running "replace:check-style" (replace) task

Running "jshint:files" (jshint) task

   node_modules/oae-util/tests/test-server.js
    171 |                    assert.ok(!me.anon)
                                                ^ Missing semicolon.
    192 |                        assert.strictEqual(me.anon, true)
                                                                  ^ Missing semicolon.

>> 2 errors in 476 files
Warning: Task "jshint:files" failed. Use --force to continue.
Error: Task "jshint:files" failed.
    at Task.<anonymous> (/Users/branden/Source/apereo/oae/Hilary/node_modules/grunt/lib/util/task.js:205:15)
    at null._onTimeout (/Users/branden/Source/apereo/oae/Hilary/node_modules/grunt/lib/util/task.js:241:33)
    at Timer.listOnTimeout [as ontimeout] (timers.js:112:15)
```

PR #1251, opened 2015-11-16
Allow multiple email domains per tenant

simong:
This is something that has come up during development and bringing together the list of 20k tenancies. We also received feedback from ESUP that some of their institutions have multiple email domains, which is currently impacting their ability to use the invitations feature.

### DIFF
--- a/etc/migration/12.1-to-12.2/2.tenantEmailDomains.js
+++ b/etc/migration/12.1-to-12.2/2.tenantEmailDomains.js
@@ -79,7 +79,7 @@ OAE.init(config, function(err) {
         var tenantsToUpdate = _.chain(rows)
             .map(Cassandra.rowToHash)
             .filter(function(hash) {
-                return (hash.emailDomain)
+                return (hash.emailDomain);
             })
             .value();
 
@@ -110,7 +110,8 @@ function _updateTenants(tenants, callback) {
     }
 
     // Update tenants in batches
-    var tenantsToUpdate = tenants.splice(0, 250);
+    var numberOfTenants = 250;
+    var tenantsToUpdate = tenants.splice(0, numberOfTenants);
     var queries = _.map(tenantsToUpdate, function(tenant) {
         return Cassandra.constructUpsertCQL('Tenant', 'alias', tenant.alias, {'emailDomains': tenant.emailDomain});
     });
@@ -119,6 +120,7 @@ function _updateTenants(tenants, callback) {
             return callback(err);
         }
 
+        log.info('Updated %d tenants their email domains', numberOfTenants);
         return _updateTenants(tenants, callback);
     });
-};
+}
--- a/etc/migration/12.1-to-12.2/2.tenantEmailDomains.js
+++ b/etc/migration/12.1-to-12.2/2.tenantEmailDomains.js
@@ -120,7 +120,7 @@ function _updateTenants(tenants, callback) {
             return callback(err);
         }
 
-        log.info('Updated %d tenants their email domains', numberOfTenants);
+        log().info('Updated %d tenants their email domains', tenantsToUpdate.length);
         return _updateTenants(tenants, callback);
     });
 }
--- a/etc/migration/12.1-to-12.2/2.tenantEmailDomains.js
+++ b/etc/migration/12.1-to-12.2/2.tenantEmailDomains.js
@@ -14,10 +14,7 @@
  */
 
 /*
- * This migration script will:
- *  - Add an `emailDomains` column to the `Tenant` column family
- *  - Copy each row's `emailDomain` value to `emailDomains`
- *  - Drop the `emailDomain` value
+ * This migration script will copy each row's `emailDomain` value to `emailDomains`
  */
 
 var _ = require('underscore');
@@ -107,7 +104,7 @@ OAE.init(config, function(err) {
  * @param  {Object}     callback.err    An error that occurred, if any
  * @api private
  */
-var _updateTenants = function(tenants, callback) {
+function _updateTenants(tenants, callback) {
     if (_.isEmpty(tenants)) {
         return callback();
     }
@@ -122,6 +119,6 @@ var _updateTenants = function(tenants, callback) {
             return callback(err);
         }
 
-        _updateTenants(tenants, callback);
+        return _updateTenants(tenants, callback);
     });
 };
--- a/etc/migration/12.1-to-12.2/2.tenantEmailDomains.js
+++ b/etc/migration/12.1-to-12.2/2.tenantEmailDomains.js
@@ -1,0 +1,127 @@
+/*!
+ * Copyright 2015 Apereo Foundation (AF) Licensed under the
+ * Educational Community License, Version 2.0 (the "License"); you may
+ * not use this file except in compliance with the License. You may
+ * obtain a copy of the License at
+ *
+ *     http://opensource.org/licenses/ECL-2.0
+ *
+ * Unless required by applicable law or agreed to in writing,
+ * software distributed under the License is distributed on an "AS IS"
+ * BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express
+ * or implied. See the License for the specific language governing
+ * permissions and limitations under the License.
+ */
+
+/*
+ * This migration script will:
+ *  - Add an `emailDomains` column to the `Tenant` column family
+ *  - Copy each row's `emailDomain` value to `emailDomains`
+ *  - Drop the `emailDomain` value
+ */
+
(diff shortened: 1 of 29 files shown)


simong:
Added a migration script and [an explanation in the upgrade guide](https://github.com/oaeproject/Hilary/wiki/OAE-Upgrade-Guide#upgrading-from-oae-121-to-oae-122).

etc/migration/12.1-to-12.2/2.tenantEmailDomains.js
```diff
@@ -0,0 +1,127 @@
...
+ * BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express
+ * or implied. See the License for the specific language governing
+ * permissions and limitations under the License.
+ */
+
+/*
+ * This migration script will:
+ *  - Add an `emailDomains` column to the `Tenant` column family
+ *  - Copy each row's `emailDomain` value to `emailDomains`
+ *  - Drop the `emailDomain` value
```
mrvisser: This script does not add an `emailDomains` column

etc/migration/12.1-to-12.2/2.tenantEmailDomains.js
```diff
@@ -0,0 +1,127 @@
...
+
+/**
+ * Update a set of tenants their email domains
+ *
+ * @param  {Object[]}   tenants         Each object represent a tenant that needs to be updated. The `alias` and `emailDomain` keys should be present
+ * @param  {Function}   callback        Standard callback function
+ * @param  {Object}     callback.err    An error that occurred, if any
+ * @api private
+ */
+var _updateTenants = function(tenants, callback) {
```
mrvisser: Since we're running in a scripted environment, can you make this `function _updateTenants` just to avoid hoisting-related oddities

etc/migration/12.1-to-12.2/2.tenantEmailDomains.js
```diff
@@ -0,0 +1,127 @@
...
+    var tenantsToUpdate = tenants.splice(0, 250);
+    var queries = _.map(tenantsToUpdate, function(tenant) {
+        return Cassandra.constructUpsertCQL('Tenant', 'alias', tenant.alias, {'emailDomains': tenant.emailDomain});
+    });
+    Cassandra.runBatchQuery(queries, function(err) {
+        if (err) {
+            return callback(err);
+        }
+
+        _updateTenants(tenants, callback);
```
mrvisser: Needs to be `return`ed

etc/migration/12.1-to-12.2/2.tenantEmailDomains.js
```diff
@@ -0,0 +1,124 @@
...
+        if (err) {
+            log().error({'err': err}, 'Unable to get all the tenants from the database');
+            return process.exit(err.code);
+        }
+
+        // Iterate through all the tenants and copy the emailDomain
+        var tenantsToUpdate = _.chain(rows)
+            .map(Cassandra.rowToHash)
+            .filter(function(hash) {
+                return (hash.emailDomain)
```
mrvisser: I think the linting needs to be run. This is missing a `;`

etc/migration/12.1-to-12.2/2.tenantEmailDomains.js
```diff
@@ -0,0 +1,124 @@
...
+    var queries = _.map(tenantsToUpdate, function(tenant) {
+        return Cassandra.constructUpsertCQL('Tenant', 'alias', tenant.alias, {'emailDomains': tenant.emailDomain});
+    });
+    Cassandra.runBatchQuery(queries, function(err) {
+        if (err) {
+            return callback(err);
+        }
+
+        return _updateTenants(tenants, callback);
+    });
```
mrvisser: Can you give some log feedback while this is running?

etc/migration/12.1-to-12.2/3.dropEmailDomain.cql
```diff
@@ -0,0 +1 @@
+ALTER TABLE "Tenant" drop "emailDomain" text;
```
mrvisser: This fails because the column type `text` isn't part of the drop syntax. Also `drop` should be upper case.

node_modules/oae-tenants/lib/api.js
```diff
@@ -800,20 +814,22 @@ var _setLandingPageBlockAttribute = function(ctx, block, blockName, attributeNam
...
+    _.each(emailDomains, function(emailDomain) {
+        var matchingTenantAlias = tenantEmailDomainIndex.conflict(updateTenantAlias, emailDomain);
+        var matchingTenant = tenants[matchingTenantAlias];
+        var matchingEmailDomains = (matchingTenant && matchingTenant.emailDomains);
+
+        var err = {
+            'code': 400,
+            'msg': util.format('The email domain "%s" conflicts with existing email domains: %s', emailDomain, matchingEmailDomains)
+        };
+        validator.check(matchingEmailDomains, err).isNull();
```
mrvisser: Should this `isNull` change to something that's more appropriate for an array that could be empty?

node_modules/oae-tenants/lib/rest.js
```diff
@@ -206,7 +207,8 @@ OAE.globalAdminRouter.on('post', '/api/tenantNetwork/:id/removeTenants', functio
  * @HttpResponse                    400             The tenant alias should not contain a space
  */
 OAE.globalAdminRouter.on('post', '/api/tenant/create', function(req, res) {
-    var opts = _.oaeExtendDefined({}, _.pick(req.body, 'emailDomain', 'countryCode'));
+    var opts = _.oaeExtendDefined({}, _.pick(req.body, 'countryCode'));
+    opts.emailDomains = OaeUtil.toArray(req.body.emailDomain);
```
mrvisser: If our tenant UI parameter is `emailDomains`, then I think the query string parameter to set its value with should also be `emailDomains`, not `emailDomain`

node_modules/oae-tenants/tests/test-tenants.js
```diff
@@ -325,6 +325,21 @@ describe('Tenants', function() {
...
+            var index = _createIndex();
+
+            index.update('sussex', 'sussex.ac.uk');
+            index.update('sussex', 'susx.ac.uk');
+
+            assert.strictEqual(index.match('sussex.ac.uk'), 'sussex');
+            assert.strictEqual(index.match('susx.ac.uk'), 'sussex');
+
+            callback();
+        });
```
mrvisser: I'd feel a bit safer with more tests based around adding, deleting and checking for conflicts for test cases with multiple email domains.

mrvisser:
Added some review, back to @simong

etc/migration/12.1-to-12.2/2.tenantEmailDomains.js
```diff
@@ -0,0 +1,126 @@
...
+    var tenantsToUpdate = tenants.splice(0, numberOfTenants);
+    var queries = _.map(tenantsToUpdate, function(tenant) {
+        return Cassandra.constructUpsertCQL('Tenant', 'alias', tenant.alias, {'emailDomains': tenant.emailDomain});
+    });
+    Cassandra.runBatchQuery(queries, function(err) {
+        if (err) {
+            return callback(err);
+        }
+
+        log.info('Updated %d tenants their email domains', numberOfTenants);
```
mrvisser: This needs to be `log().info(...`

mrvisser: Also, instead of `numberOfTenants`, it should be `tenantsToUpdate.length`

mrvisser:
It seems I can update a tenant to have a duplicate email domain when I update it to multiple domains. Steps:

1. Create 2 tenants, `foo` and `bar`
2. Give tenant `foo` the email domain `foo.ac.uk`
3. Give tenant `bar` the email domains `foo.ac.uk, bar.ac.uk`

And that seems to succeed

mrvisser:
Assigned to @simong for followup

simong:
Was that through the admin UI? I just pushed a (UI) fix that ensures it sends an array for `emailDomains` and not comma separated string.

mrvisser:
Yes it was through the admin UI, were you able to reproduce the issue?

simong:
The bug in the admin UI was causing it to send `emailDomains=["foo.test.com, bar.test.com"]` (not the single value in the array)

mrvisser:
If that's the case, Hilary should never have accepted `"[\"foo.test.com, bar.test.com\"]"` as the email domains value. Can you fix that and ensure there is a test for it?

mrvisser:
There's a linting failure. I'll fix it.

PR #1254, opened 2015-11-17
Method to know which software version an environment is running

simong:
Hi,

Please can we have some sort of api call that a non-admin can do to get the code version running in an environment?
I'm thinking something like;
  https://tenant.unity.ac/api/version
to return something like;
OAE 12.1.1

Cheers,
D

### DIFF
--- a/node_modules/oae-version/lib/api.js
+++ b/node_modules/oae-version/lib/api.js
@@ -18,6 +18,7 @@ var gift = require('gift');
 var path = require('path');
 var util = require('util');
 
+var IO = require('oae-util/lib/io');
 var log = require('oae-logger').logger('oae-version');
 
 // A variable that will hold the path to the UI directory
@@ -39,22 +40,31 @@ var init = module.exports.init = function(uiPath) {
  * Get the version information for OAE
  *
  * @param  {Function}   callback                    Standard callback function
+ * @param  {Object}     callback.err                An error that occurred, if any
  * @param  {Object}     callback.version            The version information
  * @param  {String}     callback.version.hilary     The version information for Hilary
  * @param  {String}     callback.version.3akai-ux   The version information for the UI
  */
 var getVersion = module.exports.getVersion = function(callback) {
     if (_version) {
-        return callback(_version);
+        return callback(null, _version);
     }
 
-    getHilaryVersion(function(hilaryVersion) {
-        getUIVersion(function(uiVersion) {
+    _getHilaryVersion(function(err, hilaryVersion) {
+        if (err) {
+            return callback(err);
+        }
+
+        _getUIVersion(function(err, uiVersion) {
+            if (err) {
+                return callback(err);
+            }
+
             _version = {
                 'hilary': hilaryVersion,
                 '3akai-ux': uiVersion
             };
-            return callback(_version);
+            return callback(null, _version);
         });
     });
 };
@@ -63,23 +73,25 @@ var getVersion = module.exports.getVersion = function(callback) {
  * Get the version information for the backend
  *
  * @param  {Function}   callback            Standard callback function
+ * @param  {Object}     callback.err        An error that occurred, if any
  * @param  {String}     callback.version    The version information for the backend
  * @api private
  */
-function getHilaryVersion(callback) {
+function _getHilaryVersion(callback) {
     var hilaryDirectory = path.resolve(__dirname, '..', '..', '..');
-    getVersionInfo(hilaryDirectory, callback);
+    _getVersionInfo(hilaryDirectory, callback);
 }
 
 /**
  * Get the version information for the UI
  *
  * @param  {Function}   callback            Standard callback function
+ * @param  {Object}     callback.err        An error that occurred, if any
  * @param  {String}     callback.version    The version information for the UI
  * @api private
  */
-function getUIVersion(callback) {
-    getVersionInfo(_uiPath, callback);
+function _getUIVersion(callback) {
+    _getVersionInfo(_uiPath, callback);
 }
 
 /**
@@ -90,22 +102,19 @@ function getUIVersion(callback) {
  *
  * @param  {String}     directory           The directory to get the version information for
  * @param  {Function}   callback            Standard callback function
+ * @param  {Object}     callback.err        An error that occurred, if any
  * @param  {String}     callback.version    The version information for the directory
  * @api private
  */
-function getVersionInfo(directory, callback) {
-    getBuildVersion(directory, function(err, buildVersion) {
+function _getVersionInfo(directory, callback) {
+    _hasBuildInfoFile(directory, function(err, hasBuildInfoFile) {
         if (err) {
-            return getGitVersion(directory, function(err, gitVersion) {
-                if (err) {
-                    return callback('Unknown');
-                }
-
-                return callback(gitVersion);
-            });
+            return callback(err);
+        } else if (hasBuildInfoFile) {
+            return _getBuildVersion(directory, callback);
+        } else {
+            return _getGitVersion(directory, callback);
         }
-
-        return callback(buildVersion);
     });
 }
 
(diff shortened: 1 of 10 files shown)


node_modules/oae-version/lib/api.js
```diff
@@ -0,0 +1,188 @@
...
+};
+
+/**
+ * Get the version information for the backend
+ *
+ * @param  {Function}   callback            Standard callback function
+ * @param  {String}     callback.version    The version information for the backend
+ * @api private
+ */
+function getHilaryVersion(callback) {
```
mrvisser: Non-exposed functions should be prefixed with a `_`

node_modules/oae-version/lib/api.js
```diff
@@ -0,0 +1,188 @@
...
+            return getGitVersion(directory, function(err, gitVersion) {
+                if (err) {
+                    return callback('Unknown');
+                }
+
+                return callback(gitVersion);
+            });
+        }
+
+        return callback(buildVersion);
```
mrvisser: Rather than relying on an error, we should have a function that determines if there is a build version available, then decides whether to parse a build version or use a git version instead.

Also, this is not a standard callback function as no error is ever provided as the first argument, but maybe there should be as I don't see any reason to muffle errors in this case.

If there is a build-info available, we can promote an error if it's not valid JSON, for example. It's something we'll want to be aware of and have error information for.

node_modules/oae-version/lib/api.js
```diff
@@ -0,0 +1,188 @@
...
+                'err': err,
+                'path': directory
+            }, 'Could not open the git repo to get the version information');
+            return callback({'code': 500, 'msg': 'Could not open the git repo'});
+        }
+
+        // Get the last tag info
+         repo.tags(function(err, tags) {
+            // Sort them descending (last tag first)
+            tags = tags.sort(function(a, b) {
```
mrvisser: We should handle an error here

node_modules/oae-version/lib/api.js
```diff
@@ -0,0 +1,188 @@
...
+                if (err) {
+                    log().error({
+                        'err': err,
+                        'path': directory
+                    }, 'Could not get the current git branch for the version information');
+                    return callback({'code': 500, 'msg': 'Could not get the branch information'});
+                }
+
+                var msg = 'Last tag: %s, Current branch: %s, Commit SHA1: %s, Date: %s';
+                var version = util.format(msg, lastTag.name, branch.name, branch.commit.id, branch.commit.committed_date);
```
mrvisser: Any reason not to use the same format as we use in tarballs deployed from TravisCI?

node_modules/oae-version/lib/rest.js
```diff
@@ -0,0 +1,38 @@
...
+ * @Path        /version
+ * @Return      {string}                          The currently running version
+ * @HttpResponse                    200           The version will be returned
+ */
+OAE.tenantRouter.on('get', '/api/version', getVersion);
+OAE.globalAdminRouter.on('get', '/api/version', getVersion);
+
+function getVersion(req, res) {
+    VersionAPI.getVersion(function(version) {
+        return res.status(200).send(version);
```
mrvisser: I think we should just output proper JSON like all our other endpoints

node_modules/oae-version/tests/test-version.js
```diff
@@ -0,0 +1,77 @@
...
+    });
+
+    /*!
+     * Verify the version information
+     *
+     * @param  {RestContext}        restContext     The rest context to get the version information with
+     * @param  {Function}           callback        Standard callback function
+     * @throws {AssertionError}                     Thrown if any assertions fail
+     */
+    function verifyVersionInformation(restContext, callback) {
```
mrvisser: Should start with `_`

node_modules/oae-version/tests/test-version.js
```diff
@@ -0,0 +1,77 @@
...
+        RestAPI.Version.getVersion(restContext, function(err, version) {
+            assert.ok(!err);
+            assert.ok(_.isObject(version));
```
mrvisser: Is this really supposed to be an object? Didn't we output a string from the REST endpoint? My guess is we're getting bitten by `_.isObject` working on strings :/

node_modules/oae-version/tests/test-version.js
```diff
@@ -0,0 +1,77 @@
...
+            assert.strictEqual(_.keys(version).length, 2);
```
mrvisser: You can use `_.size(version)` here to do the same.

node_modules/oae-version/lib/rest.js
mrvisser: I think we outputting proper JSON here, but the getHilaryVersion + the documentation here confused me. Can you update the docs?

node_modules/oae-version/tests/test-version.js
mrvisser: This is now invalid

node_modules/oae-version/lib/api.js
simong: There's only a version number in the tarballs, when deploying from a branch, we'll probably want a bit more information.

simong:
Pushed some follow-up

node_modules/oae-version/lib/api.js
mrvisser: We generate the version in the tarballs from git information, and we have a syntax for non-tag version, which includes the commit hash, the base tag, and how many commits we are beyond that. I think we should use it instead.

simong: I disagree, it feels like these version numbers would be most likely used in a QA environment where you quickly want to determine whether a server is on the latest version of a branch. `1.2.1+66commits+a1b2bc3d4e5` is going to be less useful than `Branch: 1254-prod-git-version, date: 2015-11-23 17:46:11`

mrvisser: > quickly want to determine whether a server is on the latest version of a branch

If this is what we want to determine, then neither approach is all that useful for someone like David who doesn't have ballpark figures of when commits were made. A more suitable approach would be to get this information specifically, such as fetching and then calculating how many commits away we are from the remote head.

> `1.2.1+66commits+a1b2bc3d4e5` is going to be less useful than `Branch: 1254-prod-git-version, date: 2015-11-23 17:46:11`

I don't think it would be substantially less useful than `1254-prod-git-version+4commits+a1b2c3` tbh. After a week goes by, exact date and times get really fuzzy.

I'm mostly trying to avoid giving an entirely different representation based on if it's on git or if it has a `build-info.json`. If we can give it a consistent consolidated model, then it's less confusing. E.g.,

```json
{
    "version": "1.2.1+66+a1b2c3",
    "date": "2015-11-23 17:46:11 UTC"
    "branch": "1254-prod-git-version"
}
```

So "version" is required and always the same format, and date+branch are optional and always the same format.

simong:
Pushed some more follow-up

mrvisser:
```
[2015-12-08T00:18:14.302Z] ERROR: oae/98948 on Brandens-MacBook-Pro.local: An uncaught exception was raised to the application
    ReferenceError: _getNumberOfCommitsSinceTag is not defined
        at /Users/branden/Source/apereo/oae/Hilary/node_modules/oae-version/lib/api.js:261:20
        at /Users/branden/Source/apereo/oae/Hilary/node_modules/gift/lib/commit.js:55:18
        at ChildProcess.exithandler (child_process.js:646:7)
        at ChildProcess.emit (events.js:98:17)
        at maybeClose (child_process.js:756:16)
        at Process.ChildProcess._handle.onexit (child_process.js:823:5)
```

node_modules/oae-version/lib/api.js
```diff
@@ -0,0 +1,289 @@
...
+ * Parse it and return the relevant version information
+ *
+ * @param  {String}     directory           The directory where the build info can be found
+ * @param  {Function}   callback            Standard callback function
+ * @param  {Object}     callback.err        An error that occurred, if any
+ * @param  {String}     callback.version    The version information for the directory
+ * @api private
+ */
+function _getBuildVersion(directory, callback) {
+    var buildInfoPath = path.join(directory, 'build-info.json');
```
mrvisser: This should use `getBuildInfoFilePath(directory)`

mrvisser:
Couple of remaining issues. Back to @simong

simong:
Pushed more follow-up

nicolaasmatthijs:
Back to @mrvisser for review

PR #1267, opened 2015-12-02
Report tenant and user name in Mixpanel user profile

simong:
Currently, the Mixpanel user profiles don't appear to contain the user's name and tenant information.


nicolaasmatthijs:
Looks good. Merged.

PR #1277, opened 2016-03-07
Inviting guests configurable per tenant

simong:
Each tenant has an “allow guests” configuration flag that determines
whether guests can be invited from this tenant. If the flag is set to
disabled, email invitations that do not match any known tenant email
domain will cause 401’s.

### DIFF
--- a/node_modules/oae-authz/lib/permissions.js
+++ b/node_modules/oae-authz/lib/permissions.js
@@ -530,19 +530,18 @@ var _validateRoleChanges = function(ctx, resource, targetRoles, opts, callback) 
     // addresses that match known tenant email domains should still get an
     // invitation. Email addresses that wind up on the guest tenant should
     // be disallowed however.
-    var guestEmails = !_.chain(emailTargetRoles)
+    var guestEmails = _.chain(emailTargetRoles)
         .pluck('email')
         .map(TenantsAPI.getTenantByEmail)
         .filter({'isGuestTenant': true})
         .value();
 
     if (!_.isEmpty(guestEmails) && !TenantsUtil.canInviteGuests(ctx.user().tenant.alias)) {
-        var err = {
+        return callback({
             'code': 401,
             'invalidPrincipals': guestEmails,
             'msg': 'Guests cannot be invited from this tenant'
-        };
-        return callback(err);
+        });
     }
 
     // Get the principal ids that were accompanied with a valid email assertion, thus bypassing
--- a/node_modules/oae-authz/lib/permissions.js
+++ b/node_modules/oae-authz/lib/permissions.js
@@ -525,6 +525,26 @@ var _validateRoleChanges = function(ctx, resource, targetRoles, opts, callback) 
     var principalTargetRoles = _.first(targetRoles);
     var emailTargetRoles = _.last(targetRoles);
 
+    // Ensure that no guests can get invited on a tenant that has disabled
+    // inviting guests. When a tenant disables inviting guests, email
+    // addresses that match known tenant email domains should still get an
+    // invitation. Email addresses that wind up on the guest tenant should
+    // be disallowed however.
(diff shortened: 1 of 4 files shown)


nicolaasmatthijs:

node_modules/oae-authz/lib/permissions.js
```diff
@@ -525,6 +525,26 @@ var _validateRoleChanges = function(ctx, resource, targetRoles, opts, callback)
...
+        .value();
+
+    if (!_.isEmpty(guestEmails) && !TenantsUtil.canInviteGuests(ctx.user().tenant.alias)) {
+        var err = {
+            'code': 401,
+            'invalidPrincipals': guestEmails,
+            'msg': 'Guests cannot be invited from this tenant'
+        };
+        return callback(err);
+    }
```
mrvisser: I think `guestEmails` is a boolean, and then we're doing a `_.isEmpty` on it?

node_modules/oae-authz/lib/permissions.js
```diff
@@ -525,6 +525,26 @@ var _validateRoleChanges = function(ctx, resource, targetRoles, opts, callback)
...
+        .filter({'isGuestTenant': true})
+        .value();
+
+    if (!_.isEmpty(guestEmails) && !TenantsUtil.canInviteGuests(ctx.user().tenant.alias)) {
+        var err = {
+            'code': 401,
+            'invalidPrincipals': guestEmails,
+            'msg': 'Guests cannot be invited from this tenant'
+        };
+        return callback(err);
```
mrvisser: We don't need to assign a variable to `err` here, we can just instantiate the object in the callback parameter. Also avoids the hazard where `err` becomes a reference in the parent function scope.

node_modules/oae-authz/tests/test-invitations.js
```diff
@@ -1859,7 +1862,25 @@ describe('Invitations', function() {
...
-                            return callback();
+                            // Ensure a user can create content and invite guests that end up on the guest tenant
+                            fns.createSucceeds(publicTenant0.publicUser.restContext, 'public', ['thisemail0@defaultstoguest.local'], [], function() {
+
+                                _disableInvitingGuests(publicTenant0.tenant.alias, function() {
+                                    // Invitations that end up on the guest tenant are disallowed
+                                    fns.createFails(publicTenant0.publicUser.restContext, 'public', ['thisemail0@defaultstoguest.local'], [], 401, function() {
+                                        // Sanity check all other invitations are still accepted
+                                        fns.createFails(privateTenant0.publicUser.restContext, 'public', [_emailForTenantInfo(publicTenant0)], [], 401, function(resource) {
+                                            fns.createFails(publicTenant0.publicUser.restContext, 'public', [_emailForTenantInfo(privateTenant0)], [], 401, function(resource) {
```
mrvisser: The above comment is incorrect for these 2 cases

mrvisser:
Couple issues I think, back to @simong

node_modules/oae-authz/lib/permissions.js
simong: No? `guestEmails` is an array of guest Tenant objects (or just the empty array)

simong: Ugh, I missed the `!`, no idea how that ended up there

mrvisser:
@simong I feel like there was definitely a bug in there regarding the `!` that test coverage wasn't catching. Should a test be added to ensure there is coverage there?

mrvisser:
back to @simong for feedback

simong:
The tests do catch this. I think the `!` snuck in there when I touched up a typo in the comment above. Travis doesn't catch this because the tests are still defaulting to `/bin/true` :(

mrvisser:
Ah, cool. Thanks!

PR #1279, opened 2016-03-08
Endpoint to lookup tenant by email address

simong:
Also added the tenant matching the query string in the email search

### DIFF
--- a/node_modules/oae-search/lib/api.js
+++ b/node_modules/oae-search/lib/api.js
@@ -157,6 +157,7 @@ var registerSearchDocumentProducer = module.exports.registerSearchDocumentProduc
  * Register a search with the Search API that will process search input and produce a query to run against elastic search. Once the query
  * is created by the search registered here, the search will be sent to ElasticSearch and and all results will be passed through the
  * document transformers. @see #registerSearchDocumentTransformer for more information.
+ * Finally, the entire result object can be processed by the search
  *
  * @param  {String}    typeName                            The name of the search. Once registered, this search can be invoked by using this `typeName` in the `search` method.
  * @param  {Function}  queryBuilder                        The search function that will be invoked when a search is performed. This function is responsible for generating a query as per the ElasticSearch Query DSL and returning it in the callback.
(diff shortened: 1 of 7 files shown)


nicolaasmatthijs:

mrvisser:
From slack:

> Hmm, I had anticipated this to just be a more descriptive error when the user tries to share w/ an invalid email address
> I’m guessing a proactive UI check is needed as a usability improvement to that?
> what I meant was that, rather than build a back-end endpoint and let the UI determine if an error should be shown, I thought we would just leave the widget the way it was, but when they sent the request to share with the emails, the error that came back would outline the emails it failed to share with
> no new back-end endpoint, no additional ajax request in the AS plugin, it would just give more information about the error when they attempt to share with them.
> but as long as we purposefully determined that there is an improvement by pro-actively making that request to check if there is an issue, I’m fine w/ it

@simong @nicolaasmatthijs can you confirm all the extra complexity is necessary?

nicolaasmatthijs:
From a usability point of view, I think it is better to do the check before the sharing actually happened. If you know before sharing that you won't be able to share with some of the people you're trying to share with, it might change your intention. If you don't know before sharing, you potentially have to start cleaning up those that did work or something along those lines.

node_modules/oae-search/lib/api.js
```diff
@@ -348,7 +348,12 @@ var search = module.exports.search = function(ctx, searchType, opts, callback) {
 
                 SearchAPI.emit(SearchConstants.events.SEARCH, ctx, searchType, opts, transformedResults);
 
-                return callback(null, transformedResults);
+                // Perform post-processing (if any)
+                var postProcessor = registeredSearch.postProcess || function(ctx, opts, results, callback) {
+                    return callback(null, results);
+                };
+
+                return postProcessor(ctx, opts, transformedResults, callback);
```
mrvisser: Can you add this to the `registerSearch` documentation?

node_modules/oae-search/lib/searches/email.js
```diff
@@ -63,3 +64,21 @@ module.exports = function(ctx, opts, callback) {
     var filter = SearchUtil.filterAnd(filterResources, filterInteractingTenants);
     return callback(null, SearchUtil.createQuery(query, filter, queryOpts));
 };
+
+/**
+ * Post process a search result set
```
mrvisser: Since it's implied that it's post processing a search result set, can this doc be more specific to what it's doing (i.e., indicating on the search result the tenant to which an email belongs)?

node_modules/oae-tenants/lib/rest.js
```diff
@@ -228,11 +228,33 @@ OAE.globalAdminRouter.on('post', '/api/tenant/create', function(req, res) {
...
+ * @Server      tenant
+ * @Method      GET
+ * @Path        /tenantsByEmail
+ * @Return      {Tenant}                        The current tenant
+ * @QueryParam  {string}        emails          The emails to match tenants on
+ * @HttpResponse                200             The tenants for each email address
+ */
+OAE.tenantRouter.on('get', '/api/tenantsByEmail', function(req, res) {
+    var emails = OaeUtil.toArray(req.query.emails);
+    if (emails.length === 0) {
```
mrvisser: `_.isEmpty(emails)`

mrvisser:
Looks good, just a few doc-related-things. Back to @simong

PR #1280, opened 2016-03-12
Shibboleth/CAS: Create user without email  address if email address is invalid

simong:
Currently Shibboleth authentication fails if the email address is invalid. We should ignore invalid email addresses and just create a user account without an email address.

### DIFF
--- a/node_modules/oae-authentication/lib/strategies/shibboleth/init.js
+++ b/node_modules/oae-authentication/lib/strategies/shibboleth/init.js
@@ -85,7 +85,17 @@ module.exports = function(config) {
             };
 
             // Get an email address from the provided headers
-            opts.email = _getBestAttributeValue(tenant.alias, 'mapEmail', headers);
+            var email = _getBestAttributeValue(tenant.alias, 'mapEmail', headers);
+
+            // Only use the email address if its valid. Don't fail the
+            // request if it's invalid
+            if (email) {
+                var validator = new Validator();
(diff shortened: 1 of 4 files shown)


simong:
Attached a commit which does this

nicolaasmatthijs:
Assigning to @mrvisser for review.

nicolaasmatthijs:
FWIW, we'd like to include this fix in next week's release as well.

node_modules/oae-authentication/lib/strategies/shibboleth/init.js
```diff
@@ -85,7 +85,17 @@ module.exports = function(config) {
...
+
+            // Only use the email address if its valid. Don't fail the
+            // request if it's invalid
+            if (email) {
+                var validator = new Validator();
+                validator.check(email, {'code': 400, 'msg': 'Invalid email'}).isEmail();
+                if (!validator.hasErrors()) {
+                    opts.email = email;
+                }
+            }
```
mrvisser: Rather than copy / paste this logic for all of these strategies, I think we can consolidate it in `AuthenticationAPI._getOrCreateUser` when `opts.authoritative === true`

mrvisser:
Left a suggestion to consolidate the validation. Back to @simong for followup

simong:
Pushed follow-up

mrvisser:
Looks good!

Issue #1284, opened 2016-04-18
Users who have not correctly populated the Name field should be hidden.


yojimbo23:
If a user signs up to the service through SSO, and the service does not provide the Name attribute, it is possible for that user to drop out of the process before updating their Name and Email address.

This can result in the creation of a user with an incomprehensible name, for example "https://registry.shibboleth.ox.ac.uk/idp!https://shib-sp.unity.ac/shibboleth!ZreQ+tXFwfhw8fLHtaPStsAcHBg="

Having these users appear causes confusion for other users of the tenancy, so they should remain hidden until they update their Name (and they should be forced to do this on next login, as appears to be the case).

nicolaasmatthijs:
Assigning to @sallakarppinen as a next ticket as it will provide some useful exposure to the auth system.

@simong @mrvisser : Feel free to add some pointers/context here.

simong:
Most of the heavy authentication lifting is done by [Passport](http://passportjs.org/). All of the relevant glue logic (or custom authentication strategies) is contained in the `oae-authentication` module.

All the different authentication strategies have their own folder under `oae-authentication/lib/strategies/`. Each strategy should export an implementation of the `oae-authentication/lib/strategy.js` interface in its `oae-authentication/lib/strategies/<strategy name>/init.js` file. An easy example to get started with is the `Twitter` strategy as it delegates almost everything to Passport:

```javascript
strategy.shouldBeEnabled = function(tenantAlias) {
        return AuthenticationConfig.getValue(tenantAlias, AuthenticationConstants.providers.TWITTER, 'enabled');
    };
```
`shouldBeEnabled` tells you whether a strategy should be enabled for a given tenant or not. Usually this just delegates to the config api

```
strategy.getPassportStrategy = function(tenant) {
```

This function should return a  [PassportJS strateg](http://passportjs.org/). Each tenant will have its own passportjs strategy object as some of them have slightly different configuration options. As you'll notice from the Twitter `getPassportStrategy` function there are some gotcha's around changing configuration values and on-the-fly reloading of authentication strategies. (if you follow the interface, this is usually not a big problem).

All passport strategies require you to specify a callback function that specifies what needs to happen once the user is authenticated. In the case of Twitter it usually means either:
 - create a new user account if this is the first time the user signed in or
 - get the existing user account and establish a new session.
[...]

PR #1285, opened 2016-04-25
Set users with incorrectly populated displayName as private

sallakarppinen:
This pull request provides a fix for issue #1284

### DIFF
--- a/node_modules/oae-authentication/lib/strategies/shibboleth/init.js
+++ b/node_modules/oae-authentication/lib/strategies/shibboleth/init.js
@@ -75,7 +75,7 @@ module.exports = function(config) {
             }
 
             // There are a lot of SAML attributes that may indicate a user's display name. The administrator
-            // should provide a suitable priority list to construct the displayname. If no suitable value was
+            // should provide a suitable priority list to construct the display name. If no suitable value was
             // returned from the mapping, we fall back to the `remote_user` attribute, as this is always provided
             var displayName = _getBestAttributeValue(tenant.alias, 'mapDisplayName', headers, headers['remote_user']);
 
@@ -83,6 +83,11 @@ module.exports = function(config) {
             var opts = {
                 'authoritative': true
             };
+
+            // Set users whose name resembles a Shibboleth identifier as private (eg. starts with http://)
+            if (displayName.indexOf('http://') === 0 || displayName.indexOf('shibboleth!') != -1) {
+                opts.visibility ='private';
+            } 
 
             // Get an email address from the provided headers
             opts.email = _getBestAttributeValue(tenant.alias, 'mapEmail', headers);
--- a/node_modules/oae-authentication/lib/strategies/shibboleth/init.js
+++ b/node_modules/oae-authentication/lib/strategies/shibboleth/init.js
@@ -84,8 +84,9 @@ module.exports = function(config) {
                 'authoritative': true
             };
 
+            var invalid = /https?:\/\/|shibboleth!|@/;
             // Set users whose name resembles a Shibboleth identifier as private (eg. starts with http://)
-            if (displayName.indexOf('http://') === 0 || displayName.indexOf('shibboleth!') != -1) {
+            if (invalid.test(displayName)) {
                 opts.visibility ='private';
             } 
 
--- a/node_modules/oae-authentication/lib/strategies/shibboleth/init.js
+++ b/node_modules/oae-authentication/lib/strategies/shibboleth/init.js
@@ -15,6 +15,7 @@
 
 var _ = require('underscore');
 
+var AuthzConstants = require('oae-authz/lib/constants').AuthzConstants;
 var Cassandra = require('oae-util/lib/cassandra');
 var ConfigAPI = require('oae-config');
 var log = require('oae-logger').logger('oae-authentication');
@@ -28,8 +29,6 @@ var AuthenticationUtil = require('oae-authentication/lib/util');
 
 var ShibbolethAPI = require('./api');
 var ShibbolethStrategy = require('./strategy');
-
-var AuthzConstants = require('oae-authz/lib/constants').AuthzConstants;
 
 module.exports = function(config) {
 
--- a/node_modules/oae-authentication/lib/strategies/shibboleth/init.js
+++ b/node_modules/oae-authentication/lib/strategies/shibboleth/init.js
@@ -29,6 +29,7 @@ var AuthenticationUtil = require('oae-authentication/lib/util');
 var ShibbolethAPI = require('./api');
 var ShibbolethStrategy = require('./strategy');
 
+var AuthzConstants = require('oae-authz/lib/constants').AuthzConstants;
 
 module.exports = function(config) {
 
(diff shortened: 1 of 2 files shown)


node_modules/oae-authentication/lib/strategies/shibboleth/init.js
```diff
@@ -84,6 +84,11 @@ module.exports = function(config) {
                 'authoritative': true
             };
 
+            // Set users whose name resembles a Shibboleth identifier as private (eg. starts with http://)
+            if (displayName.indexOf('http://') === 0 || displayName.indexOf('shibboleth!') != -1) {
```
nicolaasmatthijs: We should also catch display names that are email addresses (`@`)

node_modules/oae-authentication/lib/strategies/shibboleth/init.js
```diff
@@ -84,6 +84,11 @@ module.exports = function(config) {
...
+                opts.visibility ='private';
+            } 
+
```
nicolaasmatthijs: When the user has an invalid display name / email address, the `editprofile` widget will be displayed in the UI. At that point, the user's account visibility should not show as private, but as the default profile visibility for that tenant, so once they save they get the default visibility (unless they change it).

node_modules/oae-authentication/lib/strategies/shibboleth/init.js
nicolaasmatthijs: This should probably catch `https://` as well

node_modules/oae-authentication/tests/test-external-strategies.js
```diff
@@ -985,6 +985,53 @@ describe('Authentication', function() {
...
+                        'shib-session-id': Math.random(),
+
+                        // Fake some data about the IdP
+                        'persistent-id': 'https://idp.example.com/shibboleth#https://sp.example.com/shibboleth#' + Math.random(),
+                        'identityProvider': 'https://idp.example.com/shibboleth',
+                        'affiliation': 'Digital Services',
+                        'unscopedAffiliation': 'OAE Team',
+
+                        // Generate an external id
+                        'remote_user': 'shibboleth!' + Math.random(),
```
nicolaasmatthijs: Can we extend this to test all possible cases (`http`, `https`, `@`) ?

nicolaasmatthijs:
Finished review. Back to @sallakarppinen for follow-up

node_modules/oae-authentication/lib/strategies/shibboleth/init.js
sallakarppinen: If I understand correctly, that widget is in 3akai-ux, so do I need to make a pull request there if I need to make changes to it?

nicolaasmatthijs: That's correct

sallakarppinen: Cool - I'll have a look at how to do this and will then create a new pull request over there. Meanwhile the issues you've raised here should now all be fixed (I've extended the check to look for emails and added test cases).

node_modules/oae-authentication/lib/strategies/shibboleth/init.js
```diff
@@ -84,6 +84,12 @@ module.exports = function(config) {
                 'authoritative': true
             };
 
+            var invalid = /https?:\/\/|shibboleth!|@/i;
+            // Set users whose name resembles a Shibboleth identifier as private (eg. starts with http://)
+            if (invalid.test(displayName)) {
+                opts.visibility ='private';
```
nicolaasmatthijs: Should be using the constant `AuthzConstants.visibility.PRIVATE`

node_modules/oae-authentication/tests/test-external-strategies.js
```diff
@@ -985,6 +985,147 @@ describe('Authentication', function() {
         });
 
         /**
+         * Test that verifies that the user will be set to private if displayName resembles a shibboleth identifier
```
nicolaasmatthijs: As we're essentially using the same test 3 times, it'd be better to abstract a function that does the majority of the checks and we can just pass a remote user into. We can then execute it 4 times (for the `htttp://` case as well).

nicolaasmatthijs:
Finished review. Back to @sallakarppinen for follow-up

node_modules/oae-authentication/tests/test-external-strategies.js
```diff
@@ -984,6 +984,67 @@ describe('Authentication', function() {
             }, callback);
         });
 
+        var verifyInvalidDisplayNameMakesProfilePrivate = function(remoteUser, callback) {
```
nicolaasmatthijs: We require all functions to have JSDocs (including parameter documentation).

nicolaasmatthijs:
Left one documentation related comment. Back to @sallakarppinen

node_modules/oae-authentication/lib/strategies/shibboleth/init.js
```diff
@@ -29,6 +29,7 @@ var AuthenticationUtil = require('oae-authentication/lib/util');
 var ShibbolethAPI = require('./api');
 var ShibbolethStrategy = require('./strategy');
 
+var AuthzConstants = require('oae-authz/lib/constants').AuthzConstants;
```
nicolaasmatthijs: We follow the following ordering when requiring dependencies:

1. All external dependencies
2. All OAE dependencies from different modules
3. All OAE dependencies from the current module

Therefore, `AuthzConstants` should be added to the `var Cassandra = ...` block.

Note that we order the requires alphabetically within their block.

node_modules/oae-authentication/tests/test-external-strategies.js
```diff
@@ -985,6 +985,74 @@ describe('Authentication', function() {
         });
 
         /**
+         * Test that verifies that the user visibility will be set to private when remote_user is used to populate
```
nicolaasmatthijs: As this isn't an actual test, it should be documented as something along the lines of `Test util that ...`

node_modules/oae-authentication/tests/test-external-strategies.js
```diff
@@ -985,6 +985,74 @@ describe('Authentication', function() {
...
+                    // Fake a session ID to log in
+                    'shib-session-id': Math.random(),
+
+                    // Fake some data about the IdP
+                    'persistent-id': 'https://idp.example.com/shibboleth#https://sp.example.com/shibboleth#' + Math.random(),
+                    'identityProvider': 'https://idp.example.com/shibboleth',
+                    'affiliation': 'Digital Services',
+                    'unscopedAffiliation': 'OAE Team',
+
+                    // Generate an external id
```
nicolaasmatthijs: This comment should be updated to reflect what's happening

node_modules/oae-authentication/tests/test-external-strategies.js
```diff
@@ -985,6 +985,74 @@ describe('Authentication', function() {
...
+                        assert.strictEqual(me.visibility, 'private');
+                        callback();
+                    });
+                });
+            });
+        }
+
+        /**
+         * Test that verifies that the user visibility will be set to private if displayName resembles a shibboleth identifier,
+         * URL or email address.
```
nicolaasmatthijs: The dot at the end can be removed

nicolaasmatthijs:
Added some minor review to help highlight some of our documentation guidelines. Back to @sallakarppinen for follow-up

nicolaasmatthijs:
Looks good. Merged.

PR #1286, opened 2016-05-10
Missing email case-sensitivity migration for user profiles and email mappings

mrvisser:
Currently the `Principals` and `PrincipalsByEmail` tables for active deployments will likely have mixed case emails, depending on how they were before the 12.0 release.

There should be a migration to update it to be lower cased. This implies:

1. For all user profiles in Principals, update the email to be lower case
2. For all entries in PrincipalsByEmail, add a new row keyed by the lower-cased version with the same user id mappings, and delete the old mixed-case version

### DIFF
--- a/node_modules/oae-principals/tests/test-dao.js
+++ b/node_modules/oae-principals/tests/test-dao.js
@@ -17,6 +17,7 @@ var _ = require('underscore');
 var assert = require('assert');
 
 var PrincipalsDAO = require('oae-principals/lib/internal/dao');
+var PrincipalsTestUtil = require('oae-principals/lib/test/util');
 var RestAPI = require('oae-rest');
 var TestsUtil = require('oae-tests');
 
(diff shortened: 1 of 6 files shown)


node_modules/oae-principals/tests/test-dao.js
```diff
@@ -228,5 +217,59 @@ describe('Principals DAO', function() {
...
+                        }, function(err) {
+                            assert.ok(!err);
+                            assert.strictEqual(numPrincipalsOrig + 1, numPrincipalsAfter);
+                            assert.ok(hasNewPrincipal);
+                            return callback();
+                        });
+                    });
+                });
+            });
+        });
```
mrvisser: FYI, I just merged 2 `described` blocks that were both for `#iterateAll`. So I moved this test up into the other block.

mrvisser: Unfortunately, it butchers the diff pretty good. But if you look at the `verify updating email address retains duplicate email mappings` test on its own, that's basically all that's being added here.

mrvisser:

PR #1287, opened 2016-05-12
401 on user profile pictures in activity stream

sallakarppinen:
The following was seen for a Georgia Tech user trying to load profiles pictures from different tenants in their activity feed (comment activity):

```
"NetworkError: 401 Unauthorized - https://oae.gatech.edu/api/download/signed?uri=local%3Au%2Fgatech%2FTT%2FSv%2FRL%2Fly%2FTTSvRLlyI5%2Fprofilepictures%2F1439504094885%2Fmedium.jpg&expires=1461801600000&signature=df502326421cdc625c00387570de1e0b96f53298"
```

This caused the profile pictures to not display as expected.

screen shot 2016-05-09 at 11 54 32

### DIFF
--- a/node_modules/oae-content/tests/test-activity.js
+++ b/node_modules/oae-content/tests/test-activity.js
@@ -361,6 +361,52 @@ describe('Content Activity', function() {
         });
 
         /**
+         * Test that verifies that profile picture URLs in the comments in the activity stream are non-expiring.
+         */
+        it('verify a comment activity has a non-expiring profile picture URL', function(callback) {
+            TestsUtil.generateTestUsers(camAdminRestContext, 3, function(err, users, simong, mrvisser) {
+                assert.ok(!err);
+
+                /**
+                 * Return a profile picture stream
+                 *
+                 * @return {Stream}     A stream containing an profile picture
+                 */
+                var getPictureStream = function() {
+                    var file = __dirname + '/data/profilepic.jpg';
+                    return fs.createReadStream(file);
+                };
+
+                // Give one of the users a profile picture
+                var cropArea = {'x': 0, 'y': 0, 'width': 250, 'height': 250};
+                RestAPI.User.uploadPicture(mrvisser.restContext, mrvisser.user.id, getPictureStream, cropArea, function(err) {
+                    assert.ok(!err);
+
+                    // Create a content item to be commented on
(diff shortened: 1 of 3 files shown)


nicolaasmatthijs:

mrvisser:
@sallakarppinen bug fixes should be accompanied by a regression test, if possible. Can the problem be observed in a test, and have a unit test that ensures the conditions are resolved?

Back to @sallakarppinen for feedback.

sallakarppinen:
@mrvisser, I had a look at adding a test when writing this but couldn't find a way for it to be fitted in to existing tests... The change is in how the item is stored in Cassandra, so the test would need to check what is stored and and returned. If there is somewhere in the test suite I can do that I'm happy to add a test there, just didn't know where to look.

mrvisser:
@sallakarppinen there is an end-user issue here, so presumably there is a way to verify data that comes out of the activity feed in the tests does not exhibit the faulty behaviour. For example, can we access the activity feed and looking at the user comment entities to ensure they have non-expiring URLs?

sallakarppinen:
@mrvisser I think the activity feed in the tests is not persisted in Cassandra. Non-persisted (recent) activity feeds do have expiring URLs in them, and this doesn't pose a problem as long as that link is not what gets persisted. The issue happens when we store the user object as-is, with the expiring link, and then try to fetch it again in the future after the link has expired. This is getting a bit complicated, can we continue the discussion in slack?

sallakarppinen:
Added test.

sallakarppinen:
That error happens in the `test-backends.js` with the local backend test if I remember correctly... You should be able to comment that test out and run the rest successfully.

brecke:
I'm still getting errors but that might be due to some other cause:

```
1) Content Activity Routes verify a comment activity has a non-expiring profile picture URL:
     Uncaught TypeError: Path must be a string. Received undefined
    at assertPath (path.js:7:11)
    at Object.extname (path.js:1431:5)
    at Object.module.exports.getImageExtension (/src/brecke/Hilary/node_modules/oae-util/lib/image.js:372:20)
    at _storeCroppedPictures (/src/brecke/Hilary/node_modules/oae-principals/lib/api.picture.js:273:90)
    at /src/brecke/Hilary/node_modules/oae-principals/lib/api.picture.js:246:24
    at /src/brecke/Hilary/node_modules/oae-util/lib/tempfile.js:55:13
    at FSReqWrap.oncomplete (fs.js:123:15)
```

Feel free to merge it if it passes on your side. I'll deal with this on my side later.

PR #1260, opened 2016-05-13
Collabdoc revision HTML stored with activity

sallakarppinen:
It looks like when a collaborative document is stored in the activity streams, it's reivison HTML might be stored with it. This is resulting in queries large enough to time out.

Additionally, when the error happens, the full activity information gets logged as well resulting in an extremely large log entry.

Additionally, it is possible that it gets retried a number of times, causing critical loss in disk space in the log volume.

The above should be analyzed and fixed.

### DIFF
--- a/node_modules/oae-activity/lib/internal/transformer.js
+++ b/node_modules/oae-activity/lib/internal/transformer.js
@@ -46,14 +46,14 @@ var transformActivities = module.exports.transformActivities = function(ctx, act
             _getActivityEntitiesByObjectType(activityId, activity.target, activityEntitiesByObjectType);
         });
     } catch (err) {
-        var logActivities = [];
-        activities.forEach(function(activity) {
-            logActivities.push({
-                'activityActor': activity.actor.id,
-                'activityObject': activity.object.id,
-                'activityTarget': activity.target.id
-            });
+        var logActivities = _.map(activities, function(activity) {
+         return {
+                'actor': activity.actor.id,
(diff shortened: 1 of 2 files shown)


sallakarppinen:
Have tried to reduce the size of activity items getting into logs. Unfortunately Helenus doesn't seem to have a configuration option for retries, but my understanding is it's getting quite outdated in any case so should consider upgrading.

node_modules/oae-activity/lib/internal/transformer.js
```diff
@@ -46,7 +46,15 @@ var transformActivities = module.exports.transformActivities = function(ctx, act
...
     } catch (err) {
-        log().error({'err': err, 'activities': activities, 'user': ctx.user()}, 'Failed to get activity entities');
+        var logActivities = [];
+        activities.forEach(function(activity) {
+            logActivities.push({
+                'activityActor': activity.actor.id,
+                'activityObject': activity.object.id,
+                'activityTarget': activity.target.id
+            });
+        });
```
mrvisser: We typically use underscore for our collections operations. [`_.map`](http://underscorejs.org/#map) can be used in this case to transform the entities. Also, I don't think the keys need to be prefixed with `activity` here (e.g., `activityActor`) as we are nesting this object in an `activities` object in the log.

mrvisser:
Good find on this, back to @sallakarppinen

sallakarppinen:
Changed to using underscore.

PR #1289, opened 2016-05-24
Create feed that lists recent groups visited

sallakarppinen:
Create a new API endpoint that lists the 5 most recent groups visited by the current user. The response format of this endpoint should be a standard OAE list output that contains standard OAE group profile outputs.

This will be displayed on the personal activity page in the right hand column as a shorthand to navigate to groups you've recently visited.


nicolaasmatthijs:

PR #1273, opened 2016-05-16
Investigate whether Google Apps intermediary page can be avoided

sallakarppinen:
Currently, clicking the "Sign In" button on a tenant that's configured with Google Apps authentication causes you to be sent off to the general Google Auth page.

screen shot 2015-12-28 at 18 10 56

The user is then expected to enter an email address for that Google Apps domain, without any indication of them having to do so. When signing in with a different Google address, no error message is provided either.

We should investigate if we can:

- Send them directly to the SSO page Google forwards them to (preferred)
- Provide indication of what address they should use

Issue #809, opened 2014-02-28
Google Apps sign in policy change


nicolaasmatthijs:
Google no longer seems to support branded login pages (see http://googleappsupdates.blogspot.co.uk/2013/11/new-sign-in-page-for-google-apps.html). This causes users to see the generic Google login page, not making it clear (other than the tooltip) that it is only accessible to users from the configured Google Apps domain. The error message Google provides after logging in with a different Google account also doesn’t help at all.

We should check what we can do to make this clearer, if anything

PR #1292, opened 2016-06-02
Investigate whether Google Apps intermediary page can be avoided (2)

sallakarppinen:
This is a follow-up ticket for https://github.com/oaeproject/Hilary/pull/1273.

It seems that the intermediary page still lists all the user's accounts even if the tenant has been configured with an email domain.

This has been discussed in https://github.com/oaeproject/Hilary/issues/809 as well.

### DIFF
--- a/node_modules/oae-authentication/lib/strategies/google/rest.js
+++ b/node_modules/oae-authentication/lib/strategies/google/rest.js
@@ -15,8 +15,10 @@
 
 var passport = require('passport');
 
+var ConfigAPI = require('oae-config');
 var OAE = require('oae-util/lib/oae');
 
+var AuthenticationConfig = ConfigAPI.config('oae-authentication');
 var AuthenticationConstants = require('oae-authentication/lib/constants').AuthenticationConstants;
 var AuthenticationUtil = require('oae-authentication/lib/util');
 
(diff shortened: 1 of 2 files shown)


sallakarppinen:
I think the `hd` param was getting eaten by Passport, moved it to rest.js and can now see it locally (although still can't complete login with Google).

node_modules/oae-authentication/lib/strategies/google/rest.js
```diff
@@ -35,10 +37,17 @@ var AuthenticationUtil = require('oae-authentication/lib/util');
...
+    var domains = AuthenticationConfig.getValue(req.tenant.alias, AuthenticationConstants.providers.GOOGLE, 'domains').split(',');
+    var domain = domains.length === 1 ? domains[0] : '';
+
     var options = {
         // To avoid authenticating with the wrong Google account, we give the user the opportunity to select or add
         // the correct account during the OAuth authentication cycle
-        'prompt': 'select_account'
+        'prompt': 'select_account',
+        'hd': domain
     };
```
nicolaasmatthijs: I would probably do this after the options have been defined:

```
var options = { ... };

// If there's only one allowed domain, add that to options as the hosted domain
// @see https://developers.google.com/identity/protocols/OpenIDConnect#authenticationuriparameters
var domains = AuthenticationConfig.getValue(req.tenant.alias, AuthenticationConstants.providers.GOOGLE, 'domains').
if (domains && domains.length === 1) {
  options['hd'] = domains[0];
}
```

nicolaasmatthijs:
Back to @sallakarppinen

nicolaasmatthijs:
Looks good. Merged.

PR #1293, opened 2016-06-13
Migrate (Cambridge) users from Google auth to Shibboleth

sallakarppinen:
The Cambridge tenancy is set up with Google auth, and we would like to migrate it to Shibboleth

If we change the authentication method then we need a way to link newly created accounts (using Shib) to the one's belonging to the same users (using GA) so they still have all he same content and access.

We have established that we should be able to link the accounts using EPPN. A node.js script needs to be written to copy user emails for Cambridge from the user profile to a shibboleth login id entry associated to the user.

### DIFF
--- a/etc/migration/migrateUsersToShibboleth.js
+++ b/etc/migration/migrateUsersToShibboleth.js
@@ -94,7 +94,7 @@ function _writeErrorRow(userHash, message) {
     csvStream.write({
         'principal_id': userHash.principalId ? userHash.principalId : '',
         'email': userHash.email ? userHash.email : '',
-        'login_id': userHash.loginId ? userHash.loginId : '';,
+        'login_id': userHash.loginId ? userHash.loginId : '',
         'message': message ? message : ''
     });
     errors++;
--- a/etc/migration/migrateUsersToShibboleth.js
+++ b/etc/migration/migrateUsersToShibboleth.js
@@ -77,12 +77,27 @@ fileStream.on('error', function(err) {
     process.exit(1);
 });
 var csvStream = csv.stringify({
-    'columns': ['shib_principal_id', 'shib_email', 'shib_EPPN', 'old_principal_id', 'old_email', 'message'],
+    'columns': ['principal_id', 'email', 'message'],
     'header': true,
     'quoted': true
 });
 csvStream.pipe(fileStream);
 
+/**
+ * Write errors to a CSV file
+ *
+ * @param  {Object}     userHash                The existing user
+ * @param  {String}     message                 A short message detailing the issue
+ * @api private
+ */
+function _writeErrorRow(userHash, message) {
+    csvStream.write({
+        'principal_id': userHash.principalId ? userHash.principalId : '',
+        'email': userHash.email ? userHash.email : '',
+        'message': message ? message : ''
+    });
+    errors++;
+}
 
 // Spin up the application container. This will allow us to re-use existing APIs
 OAE.init(config, function(err) {
@@ -91,24 +106,26 @@ OAE.init(config, function(err) {
         return _exit(err.code);
     }
 
-    _getAllUsersForTenant(tenantAlias, function(err, users) {
-        _mapUsersToShibUsers(users, function(err, mappedUsers) {
-            if (err) {
-                log().error({'err':err}, 'Encountered error when migrating %s users to Shibboleth', tenantAlias);
-                return _exit(1);
-            }
+    _getAllUsersForTenant(tenantAlias, function(userHashes) {    
+        _getExternalIds(userHashes, function(usersWithIds) {
+            _mapUsersToShibLogin(usersWithIds, function(err, mappedUsers) {
+                if (err) {
+                    log().error({'err':err}, 'Encountered error when migrating %s users to Shibboleth', tenantAlias);
+                    return _exit(1);
+                }
 
-            if (_.isEmpty(mappedUsers)) {
-                util.format('No users were mapped for tenant %s', tenantAlias)
-            } else {
-                util.format('%s users were migrated to Shibboleth logins.', mappedUsers.length);
-            }
+                if (_.isEmpty(mappedUsers)) {
+                    util.format('No users were migrated for tenant %s', tenantAlias);
+                } else {
+                    util.format('%s users were migrated to Shibboleth logins.', mappedUsers.length);
+                }
 
-            if (errors > 0) {
-                log().warn('Some users were not mapped to Shibboleth users, check the CSV file for more information');
-            }
+                if (errors > 0) {
+                    log().warn('Some users could not be mapped to Shibboleth logins, check the CSV file for more information');
+                }
 
-            return _exit(0);
+                return _exit(0);
+            });
         });
     });
 });
@@ -118,24 +135,24 @@ OAE.init(config, function(err) {
  *
  * @param  {String}     tenantAlias             The tenant alias to filter by
  * @param  {Function}   callback                Invoked when users have been collected
- * @param  {Object}     callback.err            The error that occurred, if any
- * @param  {Object}     callback.userHashes     An array of user hashes
+ * @param  {Object[]}   callback.userHashes     An array of user hashes
  * @api private
  */
 var _getAllUsersForTenant = function(tenantAlias, callback) {
     var userHashes = [];
 
-    // Just get the fields we'll need later
     PrincipalsDAO.iterateAll(['principalId', 'tenantAlias', 'email'], 100, _aggregateUsers, function(err) {
         if (err) {
-            log().error({'err': err}, 'Failed to iterate all users for tenant %s', tenantAlias);
+            log().error({'err': err}, 'Failed to iterate all users');
             return _exit(1);
         }
 
         log().info('Found %s users for specified tenant', userHashes.length);
+
+        return callback(userHashes);
     });
 
-    /*
+    /*!
      * Filter users down to those that are part of the specified tenant. Then add them to the `userHashes` array
      *
      * @param  {Object[]}   principalHashes     The principals to filter and aggregate
@@ -143,7 +160,6 @@ var _getAllUsersForTenant = function(tenantAlias, callback) {
      */
     function _aggregateUsers(principalHashes, callback) {
         log().info('Checking %s principals for tenancy', principalHashes.length);
-
         _.chain(principalHashes)
             .filter(function(principalHash) {
                 return principalHash.tenantAlias === tenantAlias && PrincipalsDAO.isUser(principalHash.principalId) && principalHash.email;
@@ -155,13 +171,6 @@ var _getAllUsersForTenant = function(tenantAlias, callback) {
 
         return callback();
     }
-
-    _getExternalIds(userHashes, function(err, hashesWithId) {
-        if (err) {
-            return callback(err);
-        }
-        return callback(null, hashesWithId);
-    });
 };
 
 /**
@@ -173,25 +182,23 @@ var _getAllUsersForTenant = function(tenantAlias, callback) {
  * @api private
  */
 var _getExternalIds = function(userHashes, callback, _userHashes) {
-  _userHashes = _userHashes || [];
+    _userHashes = _userHashes || [];
 
-  if (_.isEmpty(userHashes)) {
-    return callback(null, _userHashes);
-  }
-
-  // Take the next principal hash from the collection
-  var principalHash = user.shift();
-  _findExternalId(principalHash, function(err, userHash) {
-    if (err) {
-      return callback(err);
+    if (_.isEmpty(userHashes)) {
+        return callback(_userHashes);
     }
 
-    // Accumulate the return information we want
-    _userHashes.push(userHash);
+    // Take the next principal hash from the collection
+    var principalHash = userHashes.shift();
+    _findExternalId(principalHash, function(err, userHash) {
+        if (!err && userHash) {
+            // Accumulate the return information we want
+            _userHashes.push(userHash);
+        }
 
-    // Our recursive step inside the callback for _findExternalId
-    return _getExternalIds(principalHashes, callback, _userHashes);
-  });
+        // Our recursive step inside the callback for _findExternalId
+        return _getExternalIds(userHashes, callback, _userHashes);
+    });
 };
 
 /**
@@ -208,6 +215,7 @@ function _findExternalId(userHash, callback) {
     Cassandra.runQuery('SELECT "loginId" FROM "AuthenticationUserLoginId" WHERE "userId" = ?', [userId], function(err, rows) {
         if (err) {
             log().error({'err': err}, 'Failed Cassandra query for user %s\'s loginId', userHash.email);
+            _writeErrorRow(user, 'Could not retrieve loginId for this user');
             return callback(err);
         }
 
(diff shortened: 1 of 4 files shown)


sallakarppinen:
AKA: Salla's found out about underscore. The script seems to work, but I think some accounts that are not being migrated aren't being recorded in the CSV at the moment, not sure why. Also not sure about whether the structure's right so comments on how to improve very much appreciated.

etc/migration/migrateUsersToShibboleth.js
```diff
@@ -0,0 +1,388 @@
...
+            .each(function(userHash) {
+                // Add the external ID for each user
+                _findExternalId(userHash, function(err, userHash) {
+                    if (err) {
+                        return callback(err);
+                    }
+
+                    log().info('Adding user %s to tenant users', userHash.email);
+                    userHashes.push(userHash);
+                });
```
mrvisser: This here is a bit of an anti-pattern, and can cause unreliable results with respect to asynchronous processing and race conditions.

Essentially what you're doing is, for each principal in the `principalHashes`, executing a query that does not wait for the return. This can potentially result in thousands of in-flight requests waiting for response from the database. Notice how the `.each` doesn't have a callback to indicate when to continue processing, that's kind of what tips off what's happening here.

What you might want to try instead is a recursive function that starts with the array of principals hashes, and executes `_findExternalId` for each principal, then invoking the next recursive step after the external id has been accessed for each principal. Here's a rough example of looping over asynchronous callbacks:

```javascript
function _aggregateUsers(principalHashes, callback, _userHashes) {
  _userHashes = _userHashes || []
  if (_.isEmpty(principalHashes)) {
    // This is our base case, when there are no principal hashes left
    return callback(null, principalHashes);
  }

  // Take the next principal hash from the collection
  var principalHash = principalHashes.shift();
  _findExternalId(principalHash, function(err, userHash) {
    if (err) {
      return callback(err);
    }

    // Accumulate the return information we want
    _userHashes.push(userHash);

    // Our recursive step inside the callback for _findExternalId
    return _aggregateUsers(principalHashes, callback, userHashes)
  });
}
```

There are tools like `async` that makes these patterns a bit easier as well, however we don't really use it yet.

This pattern could cause some of the strange results you've been seeing.

etc/migration/migrateUsersToShibboleth.js
```diff
@@ -0,0 +1,388 @@
...
+        _exit(0);
+    }
+
+    log().info('Found %s users with Shibboleth accounts', shibbolethUsers.length);
+
+    var nonShibUsers = _.difference(allUsers, shibbolethUsers);
+
+    var mappedUsers = [];
+
+    _.each(shibbolethUsers, function(userHash) {
```
mrvisser: This is also exhibiting the unreliable asynchronous loop pattern.

etc/migration/migrateUsersToShibboleth.js
```diff
@@ -0,0 +1,388 @@
...
+                    } else {
+                        mappedUsers.push(userMatch);
+                    }
+
+                    if ((shibbolethUsers.length -1) === shibbolethUsers.indexOf(userHash)) {
+                        _.each(_.difference(nonShibUsers, mappedUsers), function(user) {
+                            _writeErrorRow(null, user, util.format('No matching Shibboleth user found for user %s', user.email));
+                        });
+
+                        callback(null, mappedUsers);
```
mrvisser: Note that in this pattern, the callback passed to `_mapToShibUsers` will get invoked once for each user passed into the function. When switching to a recursive asynchronous loop, we'll want to verify that the callback only gets invoked once.

etc/migration/migrateUsersToShibboleth.js
```diff
@@ -0,0 +1,388 @@
...
+
+/**
+ * Write errors to a CSV file
+ *
+ * @param  {Object}     shibUser                The Shibboleth user involved
+ * @param  {Object}     oldUser                 The existing user we want to map to
+ * @param  {String}     message                 A short message detailing the issue
+ * @api private
+ */
+var _writeErrorRow = function(shibUser, oldUser, message) {
```
mrvisser: This can be changed to `function _writeErrorRow`. There is a subtle yet important difference between `var _writeErrorRow = function(...` and `function _writeErrorRow(...`.

* When using `var`, the `_writeErrorRow` symbol will be available anywhere in the scope that it's declared, however its value will be `undefined` until this area of the code is evaluated. When doing lots of asynchronous operations, that typically isn't an issue, however if you try and invoke `_writeErrorRow` at the top of this file before you hit an async function, you'll see you have an undefined variable.
* When doing `function`, you'll find that the symbol is available at the beginning of the file *and* it's definition is set to the function, so running `_writeErrorRow(myUser, oldUser, 'oops!')` at the top of this file will work as you'd expect

etc/migration/migrateUsersToShibboleth.js
```diff
@@ -0,0 +1,388 @@
...
+            log().error({'err': err}, 'Failed Cassandra query for Shibboleth metadata');
+            return callback(err);
+        }
+
+        // Get the EPPN value from the Shibboleth metadata (stored as JSON)
+        var eppn = _.propertyOf(JSON.parse(rows[0].get('allAttributes').value))('eppn');
+
+        return callback(null, eppn);
+    });
+}
```
mrvisser: In the migration case where users haven't authenticated with shibboleth yet, we won't have any shibboleth metadata available. Thankfully, the shibboleth `eppn` is set to the email address of the user, so we can copy over the existing gmail address associated to their gmail authentication login id entry as their equivalent shibboleth login id.

mrvisser:
So, if I understand correctly, we don't currently have a population of cambridge shibboleth accounts in our database. We're preparing to enable cambridge shibboleth in such a way that we won't have a bunch of duplicate accounts created.

If that's correct, the portions of this PR where we are trying to delete a cambridge shibboleth user isn't really necessary.

I think all we need to do is make the new `AuthenticationLoginId` and `AuthenticationUserLoginId` entries that allow a user who logs in from shibboleth to be logged in as their existing account based on the gmail authentication.

To do that, we are making the validated assumption that the `eppn` of users coming in from cambridge shibboleth is identical to the login id that is provided by gmail (i.e., their email address). Therefore, for each user we can look up their gmail address entry in `AuthenticationUserLoginId`, then add an equivalent `cam:shibboleth:mrvisser@cam.ac.uk` entry, then add the appropriate reverse mapping in `AuthenticationLoginId` by associating `cam:shibboleth:mrvisser@cam.ac.uk` with their user id.

I don't think it's even necessary to delete the existing google login id, as if anything goes wrong, it allows us to re-enable google and the google login ids are still working as they were before.

etc/migration/migrateUsersToShibboleth.js
sallakarppinen: I'm not sure if that will work - I think it depends on which Shib parameter they've chosen as the best match (and I think that's `persistentId` not `eppn`)?

sallakarppinen:
I've added a commit with just structural changes trying to fix the broken pattern identified by @mrvisser above. I'll have a look at changing the logic to handle the case where no Shibboleth accounts have yet been created next. However, @mrvisser, could you have a look and let me know if I'm on the right track?

etc/migration/migrateUsersToShibboleth.js
mrvisser: There is an assumption that when shib is enabled, we will configure them to use `eppn` as the identification parameter instead of `persistentId` so that it maps to the same value being used by google.

sallakarppinen: I'll talk to Jim when he's back tomorrow and check if Cambridge have expressed any strong preference on what parameter they want to use.

mrvisser: This was a detail that we were already settled on in a previous email chain (that I hope was forwarded to you when asked to work on this :P).

If offering a choice to Cambridge, we should keep in mind that if they want to use persistent-id, it will not be possible to migrate accounts as we don't get a persistent-id until after a user logs in for the first time.

If some other attribute is suggested (e.g., some other kind of internal id they have), then we will need a full mapping from Cambridge that maps those new internal ids to each user's email address, or else we will not be able to perform a migration.

yojimbo23:
_So, if I understand correctly, we don't currently have a population of cambridge shibboleth accounts in our database. We're preparing to enable cambridge shibboleth in such a way that we won't have a bunch of duplicate accounts created.

If that's correct, the portions of this PR where we are trying to delete a cambridge shibboleth user isn't really necessary._

**This is correct.**

_I think all we need to do is make the new AuthenticationLoginId and AuthenticationUserLoginId entries that allow a user who logs in from shibboleth to be logged in as their existing account based on the gmail authentication.

To do that, we are making the validated assumption that the eppn of users coming in from cambridge shibboleth is identical to the login id that is provided by gmail (i.e., their email address). Therefore, for each user we can look up their gmail address entry in AuthenticationUserLoginId, then add an equivalent cam:shibboleth:mrvisser@cam.ac.uk entry, then add the appropriate reverse mapping inAuthenticationLoginId by associating cam:shibboleth:mrvisser@cam.ac.uk with their user id._


**As per the reference to the email thread above we have established as far as possible that this is the case. We should be using eppn as the identification parameter**

_I don't think it's even necessary to delete the existing google login id, as if anything goes wrong, it allows us to re-enable google and the google login ids are still working as they were before._

**Sounds sensible.**

sallakarppinen:
Script now creates EPPN-based loginId records for all users with non-duplicated emails.

etc/migration/migrateUsersToShibboleth.js
```diff
@@ -0,0 +1,352 @@
...
+        })
+        .map(function(grouped) {
+            if (grouped.length === 1) {
+                return grouped[0];
+            } else {
+                _.each(grouped, function(dupe) {
+                    _writeErrorRow(dupe, 'Found more than one user with this email');
+                });
+            }
+        })
```
mrvisser: Why are those with duplicate email addresses being filtered out?

etc/migration/migrateUsersToShibboleth.js
```diff
@@ -0,0 +1,352 @@
...
+                if (err) {
+                    log().error({'err':err}, 'Encountered error when migrating %s users to Shibboleth', tenantAlias);
+                    return _exit(1);
+                }
+
+                if (_.isEmpty(mappedUsers)) {
+                    util.format('No users were migrated for tenant %s', tenantAlias);
+                } else {
+                    util.format('%s users were migrated to Shibboleth logins.', mappedUsers.length);
+                }
```
mrvisser: I think you meant to log these

etc/migration/migrateUsersToShibboleth.js
```diff
@@ -0,0 +1,352 @@
...
+     * Filter users down to those that are part of the specified tenant. Then add them to the `userHashes` array
+     *
+     * @param  {Object[]}   principalHashes     The principals to filter and aggregate
+     * @param  {Function}   callback            Will be invoked when the principals are aggregated
+     */
+    function _aggregateUsers(principalHashes, callback) {
+        log().info('Checking %s principals for tenancy', principalHashes.length);
+        _.chain(principalHashes)
+            .filter(function(principalHash) {
+                return principalHash.tenantAlias === tenantAlias && PrincipalsDAO.isUser(principalHash.principalId) && principalHash.email;
```
mrvisser: Any reason to filter to only users who have an email address set?

etc/migration/migrateUsersToShibboleth.js
```diff
@@ -0,0 +1,352 @@
...
+            })
+            .each(function(userHash) {
+                log().info('Adding user %s to tenant users', userHash.email);
+                userHashes.push(userHash);
+            });
+
+        return callback();
```
mrvisser: This function doesn't need a callback as it is only doing synchronous work.

mrvisser: Nevermind, I'm wrong. It's part of the `iterateAll` contract.

etc/migration/migrateUsersToShibboleth.js
```diff
@@ -0,0 +1,352 @@
...
+        return callback(_userHashes);
+    }
+
+    // Take the next principal hash from the collection
+    var principalHash = userHashes.shift();
+    _findExternalId(principalHash, function(err, userHash) {
+        if (!err && userHash) {
+            // Accumulate the return information we want
+            _userHashes.push(userHash);
+        }
```
mrvisser: What are the potential errors for this callback, and why are we choosing to bypass the error without existing the script?

etc/migration/migrateUsersToShibboleth.js
sallakarppinen: Because we don't know which email we should map? Plus I assumed that if we added two identical `loginId`s into the AuthenticationLoginId table, Cassandra would be unhappy.

etc/migration/migrateUsersToShibboleth.js
sallakarppinen: We're matching on email address so if they don't have one we can't map them.

etc/migration/migrateUsersToShibboleth.js
```diff
@@ -0,0 +1,352 @@
...
+            log().error({'err': err}, 'Failed Cassandra query for user %s\'s loginId', userHash.email);
+            _writeErrorRow(user, 'Could not retrieve loginId for this user');
+            return callback(err);
+        }
+
+        var loginId = _.chain(rows)
+            .map(Cassandra.rowToHash)
+            .pluck('loginId')
+            .first()
+            .value();
```
mrvisser: Do we have any users in the cam tenant that have multiple login ids? We want to make sure we get specifically the google one, I think.

etc/migration/migrateUsersToShibboleth.js
```diff
@@ -0,0 +1,352 @@
...
+                    _writeErrorRow(dupe, 'Found more than one user with this email');
+                });
+            }
+        })
+        .compact()
+        .value();
+
+    if (_.isEmpty(cleanedUsers)) {
+        log().info('No suitable users found for tenant %s', tenantAlias);
+        callback(null, null);
```
mrvisser: This callback isn't being returned.

etc/migration/migrateUsersToShibboleth.js
```diff
@@ -0,0 +1,352 @@
...
+    }
+
+    _createShibLoginRecords(cleanedUsers, function(err, mappedUsers) {
+        if (err) {
+            callback(err);
+        }
+
+        callback(null, mappedUsers);
```
mrvisser: These callbacks here aren't being returned.

etc/migration/migrateUsersToShibboleth.js
```diff
@@ -0,0 +1,352 @@
...
+    }
+
+    var userHash = users.shift();
+
+    log().info('Processing user %s', userHash.email);
+
+    // Create a new login record for user
+    _createNewUserLogin(userHash, function(err) {
+        if (err) {
+            callback(err);
```
mrvisser: This callback isn't returned.

etc/migration/migrateUsersToShibboleth.js
sallakarppinen: Yup, I did. Will fix.

etc/migration/migrateUsersToShibboleth.js
sallakarppinen: Basically, an error when trying to retrieve data from Cassandra. We're logging it in `_findExternalId` method and  adding the user to the CSV, and I thought since it'd mostly likely be data error we didn't need to terminate.

etc/migration/migrateUsersToShibboleth.js
```diff
@@ -0,0 +1,352 @@
...
+ *
+ * @param  {Object}     userHash                The user account we want to keep
+ * @param  {Function}   callback                Standard callback function
+ * @param  {Object}     callback.err            An error that occurred, if any
+ * @api private
+ */
+var _createNewUserLogin = function(userHash, callback) {
+    var email = userHash.email;
+    var userId = userHash.principalId;
+    var newLoginId = util.format('%s:shibboleth:%s', tenantAlias, email);
```
mrvisser: A user can change their profile email to something that is no longer the original eppn. Since we want google login ids only to be remapped as shibboleth ids with the same external id, we can directly parse the email address portion of the google login ids and add a new entry with the `cam:shibboleth:` prefix.

So if I logged in with gmail, I would have a login id: `cam:google:mrvisser@cam.ac.uk`. When doing the mapping, we can parse out the `mrvisser@cam.ac.uk` and then add a new entry `cam:shibboleth:mrvisser@cam.ac.uk`.

mrvisser:
I've added some feedback, looking good! Assigning back to @sallakarppinen

etc/migration/migrateUsersToShibboleth.js
yojimbo23: There are definitely users with multiple accounts. I exported a recent list here https://research.unity.ac/content/rr/VytxgFlkZ In reality I think these may all be dev accounts of various kinds.

sallakarppinen: Should not, since they only have Google auth enabled. That's assuming they've never allowed local login, but can't check until staging is back up (I don't want to go run queries on production).

etc/migration/migrateUsersToShibboleth.js
yojimbo23: I presume the script will output these somewhere so we can follow up manually. In reality I think that if we follow the suggestion below to directly parse the email address portion of the google login ids these would be extremely unlikely to occur.

etc/migration/migrateUsersToShibboleth.js
mrvisser: Yea, good call, it would cause weird results. However, since we'll be keying off the eppn of the google login id rather than the profile email, I don't think we have to consider profile emails anymore.

sallakarppinen:
Changed to take email value from Google `loginId` plus fixes based on comments.

etc/migration/migrateUsersToShibboleth.js
```diff
@@ -0,0 +1,346 @@
...
+        .filter(function(user) {
+            var valid = user.loginId;
+            if (!valid) {
+                // We'll also end up here if they have a non-Google loginId
+                _writeErrorRow(user, 'This user has an invalid loginId');
+            }
+            return valid;
+        })
+        .compact()
+        .value();
```
mrvisser: I think this block needs to be revised now that we aren't considering duplicate email. The comment is out of date, and some steps like compacting I don't believe is necessary.

etc/migration/migrateUsersToShibboleth.js
```diff
@@ -0,0 +1,346 @@
...
+/**
+ * Create new authentication records linked to Shibboleth EPPN for all users within the tenant
+ *
+ * @param  {Object[]}   users                   The users to map
+ * @param  {Function}   callback                Standard callback function
+ * @param  {Object}     callback.err            An error that occurred, if any
+ * @param  {Object[]}   callback.mappedUsers    The Shibboleth users that were mapped to existing users
+ * @api private
+ */
+var _createShibLoginRecords = function(users, callback, _mappedUsers) {
```
mrvisser: The use of variable naming around `users` v.s. `userHashes` is inconsistent making it a little tough to follow. Can you use `userHash` when referring to a raw object fetched from cassandra (which I believe is always the case in this file). The term `user` for a variable is typically used when we have a `User` object.

mrvisser:
I've finished a round of review. Now that we have the coding patterns down path here, it would be good to see some unit tests for the migration. A sample of other migration unit tests can be found in `test-migration.js`. Something along the lines of generating 2 tenants with a good amount of users in it (say, 500 each), then running the migration on one of the tenants and ensuring that the new tenant ids are migrated and the old tenant still has their login id information intact.

sallakarppinen:
@mrvisser: I can't seem to find that test file, can you point me to it please?

mrvisser:
@sallakarppinen https://github.com/oaeproject/Hilary/blob/master/node_modules/oae-principals/tests/test-migrations.js

sallakarppinen:
In preparation for the tests, I've split the script into two files (one to run the script and interact with the filesystem, other to actually do the migration). @mrvisser, could you have a quick look and comment when you have a minute?

PR #1295, opened 2016-07-12
Searching private tenant should only ever return results from that tenant

sallakarppinen:
A private tenant (e.g., The Fold) currently returns search results that are outside of that tenant. All searches on a private tenant should be limited to that tenant. The "this tenant only" checkbox in the UI should also be hidden for private tenants.

screen shot 2016-07-06 at 06 17 46

### DIFF
--- a/node_modules/oae-search/lib/constants.js
+++ b/node_modules/oae-search/lib/constants.js
@@ -41,8 +41,8 @@ SearchConstants.general = {
     'RESOURCE_TYPE_GROUP': 'group',
     'RESOURCE_TYPE_USER': 'user',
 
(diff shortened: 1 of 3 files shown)


nicolaasmatthijs:
@sallakarppinen : This behavior should be verified in a set of comprehensive test that cover all of these different scenarios.

nicolaasmatthijs:
Back to @sallakarppinen

sallakarppinen:
@nicolaasmatthijs, I've modified the already existing tests to confirm the new behaviour (meaning search scope 'all' now functions like the network scope for non-admin users).

nicolaasmatthijs:

PR #1291, opened 2016-07-29
Provide alternative for video transcoding and thumbnailing

sallakarppinen:
As MediaCore is not actively maintained and will be disappearing soon, we should provide an alternative. The most promising route I've been able to identify is an integration with https://aws.amazon.com/elastictranscoder/, but I'd be open to other suggestions as well.

---
This change is [Reviewable](https://reviewable.io/reviews/oaeproject/hilary/1291)


sallakarppinen:
There's still work to be done here, but made into a pull-request since the next steps are going to be in 3akai-ux I think and would like some comments on the code written so far... The main issue that's not yet solved is creating signed URLs for S3, but I think the main functionality is there.

nicolaasmatthijs:
Assigning to @mrvisser for initial round of review

config.js
```diff
@@ -457,3 +457,35 @@ config.mixpanel = {
...
+ * @param  {String}     amazons3-access-key         The S3 bucket access key
+ * @param  {String}     amazons3-secret-key         The S3 bucket secret key
+ * @param  {String}     amazons3-region             The S3 bucket region
+*/
+config.storage = {
+    'amazons3-bucket': '',
+    'amazons3-region': '',
+    'amazons3-access-key': '',
+    'amazons3-secret-key': ''
+};
```
mrvisser: If this storage key only differs from our `files` storage key in that it's used for video previews instead of general user files, then should it be scoped under `config.zencoder`?

node_modules/oae-content/lib/internal/util.js
```diff
@@ -104,10 +106,35 @@ var augmentContent = module.exports.augmentContent = function(ctx, content, dura
...
+        if (content.previews.video) {
+            content.previews.video = _getSignedVideoUrl(ctx, content.previews.key);
+        }
     }
 };
 
 /**
+ * Generate a signed url that can be used to show video previews in the browser for a video file stored in AWS S3
+ *
+ * @param  {Context}    ctx         Standard context object containing the current user and the current tenant
```
mrvisser: FYI, more up-to-date wording on this parameter description is just "Current request context". Less to type, less to read, less redundant, and less half truths :P

node_modules/oae-content/lib/internal/util.js
```diff
@@ -104,10 +106,35 @@ var augmentContent = module.exports.augmentContent = function(ctx, content, dura
...
+
+    AWS.config.update({
+        'accessKeyId': config['amazons3-access-key'],
+        'secretAccessKey': config['amazons3-secret-key'],
+        'region': config['amazons3-region']
+    });
+
+    var s3 = new AWS.S3();
+
+    return s3.getSignedUrl('getObject', { Bucket: config['amazons3-bucket'], Key: key, Expiration: TIME_1_WEEK_IN_SECONDS });
```
mrvisser: Object keys need single quotes and should be split on multiple lines.

node_modules/oae-content/lib/internal/util.js
```diff
@@ -104,10 +106,35 @@ var augmentContent = module.exports.augmentContent = function(ctx, content, dura
...
+ * @return {String}                 The signed S3 url that can be used to view the video
+ */
+var _getSignedVideoUrl = function(ctx, key) {
+    var config = ZencoderUtil.getConfig(ctx.tenant().alias);
+
+    AWS.config.update({
+        'accessKeyId': config['amazons3-access-key'],
+        'secretAccessKey': config['amazons3-secret-key'],
+        'region': config['amazons3-region']
+    });
```
mrvisser: What are the "multi-threading" implications of using the AWS config in this manner? If we use the AWS config library elsewhere in the application, can there be race condition issues with updating the global config here, then it getting re-updated before some particular operation happens?

node_modules/oae-preview-processor/lib/api.js
```diff
@@ -507,11 +507,13 @@ var _generatePreviews = function(ctx, callback) {
...
         } else if (processor) {
-            processor.generatePreviews(ctx, ctx.content, function(err, ignored) {
+            processor.generatePreviews(ctx, ctx.content, function(err, ignored, pending) {
                 if (err) {
                     ctx.setStatus('error', callback);
                 } else if (ignored) {
                     ctx.setStatus('ignored', callback);
+                } else if (pending) {
+                    ctx.setStatus('pending', callback);
                 } else {
```
mrvisser: Can you extract these `'error'`, `'ignored'` and `'pending'` into an enumeration-like thing in a `constants.js` file? That way usages can be better tracked for some analysis when adding new ones like this.

mrvisser: Actually it may already be in `node_modules/oae-content/lib/constants.js` but not being used here. Can you update this to use them?

Note that test references don't have to be refactored for it.

node_modules/oae-preview-processor/lib/processors/link/vimeo.js
```diff
@@ -62,7 +62,7 @@ var generatePreviews = module.exports.generatePreviews = function(ctx, contentOb
 
         // Ignoring this video if it has no thumbnail.
         if (_.isEmpty(info) || !info[0].thumbnail_medium) {
-            return callback(null, false);
+            return callback(null, true);
```
mrvisser: Why did this change?

node_modules/oae-zencoder/lib/api.js
```diff
@@ -0,0 +1,137 @@
+/*!
+ * Copyright 2014 Apereo Foundation (AF) Licensed under the
```
mrvisser: Copyright years should be updated to 2016 for new files.

node_modules/oae-zencoder/lib/api.js
```diff
@@ -0,0 +1,137 @@
...
+        });
+    }
+
+    else if (job.state === 'failed') {
+        return callback({'err': notification, 'msg': 'Zencoder could not transcode this video for preview'});
+    }
+
+    else {
+        return callback({'err': notification, 'msg': 'Could not process notification from Zencoder'});
+    }
```
mrvisser: A couple things about this code workflow:

1. `else-` calls should be on the same line as the preceding `}`
2. We try and handle small error conditions first (e.g., `if (err) return callback(err)` in many spots). To continue with that pattern, we should first handle the `failed` and `if (job.state !== 'finished')` first
3. We try and avoid asynchronous branching inside of `if` statements. The call to `ZencoderDAO.getContentRevisionId` is an example violation of that. If we handle the 2 error cases first with returns on the callbacks, then we can run the "happy" path of this function outside of any `if` branch, which is a bit easier to read and avoids 1 unneeded level of indentation

node_modules/oae-zencoder/lib/api.js
```diff
@@ -0,0 +1,137 @@
...
+            }
+
+            log().info('Encoding of video content with the contentId %s finished successfully.', contentRevision.contentId);
+
+            _createAndSavePreviews(contentRevision, output, function(err) {
+                if (err) {
+                    return callback(err);
+                }
+
+                return callback();
```
mrvisser: In cases where we are just promoting an error and doing nothing else, it can be written like:

`return _createAndSavePreviews(contentRevision, output, callback);`

node_modules/oae-zencoder/lib/api.js
```diff
@@ -0,0 +1,137 @@
...
+    var ctx = new PreviewContext(PreviewProcessorAPI.getConfiguration(), contentRevision.contentId, contentRevision.revisionId);
+
+    ctx.login(function(err) {
+        if (err) {
+            return callback(err);
+        }
+
+        ctx.getContentData(function(err) {
+            if (ctx.revision.previews.total > 0) {
+                return callback('Previews have already been created for this revision');
```
mrvisser: This error should be a `{'code': n, 'msg': 'Previews have already...'}` style of error.

node_modules/oae-zencoder/lib/api.js
```diff
@@ -0,0 +1,137 @@
...
+            }
+
+            ctx.addPreviewMetadata('video', util.format('remote:%s', output.url));
+            ctx.addPreviewMetadata('key', output['pass_through']);
+
+            var thumbnail = _.findWhere(output.thumbnails, {label: 'base'});
+        
+            PreviewUtil.downloadRemoteFile(thumbnail.images[0].url, util.format('%s/base.png', ctx.baseDir), function(err, path) {
```
mrvisser: Should we do any handling for cases where `output.thumbnails` might not have all of these elements we're expecting? (e.g., doesn't have a thumbnail with label = base, that thumbnail doesn't have an images array, the images array is empty, it doesn't have `url`, etc...

Also, to get the first item we normally do `_.first(...)` which is a bit safer to avoid exceptions.

node_modules/oae-zencoder/lib/api.js
```diff
@@ -0,0 +1,137 @@
...
+
+        ctx.getContentData(function(err) {
+            if (ctx.revision.previews.total > 0) {
+                return callback('Previews have already been created for this revision');
+            }
+
+            ctx.addPreviewMetadata('video', util.format('remote:%s', output.url));
+            ctx.addPreviewMetadata('key', output['pass_through']);
+
+            var thumbnail = _.findWhere(output.thumbnails, {label: 'base'});
```
mrvisser: `label` key needs to be in single quotes

node_modules/oae-zencoder/lib/api.js
```diff
@@ -0,0 +1,137 @@
...
+                if (err) {
+                    return callback(err);
+                }
+
+                PreviewUtil.generatePreviewsFromImage(ctx, path, {'removeInput': true}, function(err) {
+                    if (err) {
+                        return callback(err);
+                    }
+
+                    ctx.savePreviews(function(err) {
```
mrvisser: `return ctx.savePreviews(callback);`

node_modules/oae-zencoder/lib/init.js
```diff
@@ -13,28 +13,38 @@
...
+
 module.exports = function(config, callback) {
-    // Register the mediacore video processor
-    PreviewProcessorAPI.registerProcessor('oae-file-mediacore', MediaCoreProcessor);
+    var _config = config.zencoder;
+    _.extend(_config, {'amazons3-bucket': config.storage['amazons3-bucket'],
+                       'amazons3-region': config.storage['amazons3-region'],
+                       'amazons3-access-key': config.storage['amazons3-access-key'],
+                       'amazons3-secret-key': config.storage['amazons3-secret-key']
+            });
```
mrvisser: This form of indentation is not how we align parameters, it needs to be:

```javascript
_.extend(_config, {
  'amazons3-bucket': config.storage['amazons3-bucket'],
  'amazons3-access-key': ...
});
```

mrvisser: That said, taking storage configuration from one config location and just overwriting siblings on another config object doesn't seem very straightforward. I think if we make the storage configuration scoped in the zencoder config, then it feels like it's more what we're trying to do here.

node_modules/oae-zencoder/lib/processor.js
```diff
@@ -0,0 +1,130 @@
...
+var ContentConstants = require('oae-content/lib/constants').ContentConstants;
+var Context = require('oae-context').Context;
+var log = require('oae-logger').logger('oae-zencoder');
+var PreviewConstants = require('oae-preview-processor/lib/constants');
+
+var ZencoderDAO = require('./internal/dao');
+var ZencoderUtil = require('./util');
+
+/**
+ * @borrows Interface.test as Videos.test
```
mrvisser: Should this be `... as Zencoder.test` ?

node_modules/oae-zencoder/lib/processor.js
```diff
@@ -0,0 +1,130 @@
...
+        return callback();
+    }
+
+    log().info('Generating previews for video file: %s', content.id);
+    var tenant = content.tenant;
+
+    var downloadPath = ContentUtil.getSignedDownloadUrl(new Context(tenant), revision.uri, 6000, 6000);
+
+    // Create a Zencoder job
+    ZencoderUtil.job(tenant.alias, downloadPath, content.id, ctx.revisionId, _isVideo(revision.mime), function(err, res, body) {
```
mrvisser: Rather than providing a boolean `_isVideo` to ZencoderUtil here, can we provide the mime and let ZencoderUtil do it's thing? We'll then need to expose `isVideo` and `isAudio` to the util, so we can probably have `ZencoderUtil.isVideo(mimeType)` as a function.

node_modules/oae-zencoder/lib/processor.js
```diff
@@ -0,0 +1,130 @@
...
+        if (_checkError(err, 'Error POSTing the video to Zencoder for encoding', res, body, callback)) {
+            return callback(err);
```
mrvisser: This pattern of passing `callback` into `checkError` and also potentially promoting an error smells fishy. If `_checkError` is really going to potentially invoke the callback, it should probably handle the `return callback(err)` case as well instead of returning a truthy value to trigger it.

node_modules/oae-zencoder/lib/processor.js
```diff
@@ -0,0 +1,130 @@
...
+        }
+
+        var output = body.outputs[0];
```
mrvisser: Should we do any null checking here?

node_modules/oae-zencoder/lib/processor.js
```diff
@@ -0,0 +1,130 @@
...
+            if (err) {
+                return callback(err);
+            }
+
+            ctx.setStatus(ContentConstants.previews.PENDING, function(err) {
+                if (err) {
+                    return callback(err);
+                }
+
+                return callback(null, false, true);
```
mrvisser: Rather than provide booleans to flag status, I think we should just pass the status back. This API as-is allows for a state of `return callback(null, true, true)` which doesn't make sense.

node_modules/oae-zencoder/lib/processor.js
```diff
@@ -0,0 +1,130 @@
...
+/**
+ * Determine whether or not the given mimeType is that of an audio
+ *
+ * @param  {String}     mimeType        The mimetype to test
+ * @return {Boolean}                    `true` or `false`, indicating whether or not the given mime type indicates an audio file
+ * @api private
+ */
+var _isAudio = function(mimeType) {
+    return (PreviewConstants.TYPES.AUDIO.indexOf(mimeType) !== -1);
+};
```
mrvisser: Rather than have `ZencoderUtil.isVideo` etc... like I mentioned before, since we already have some recognition of audio/video in the core preview constants, we could even have `PreviewUtil.isVideo(mimeType)` etc... which then could be re-used for other audio/video preview modules.

mrvisser: Also, is there a `_.contains` ? I think that would be preferred.

node_modules/oae-zencoder/lib/processor.js
```diff
@@ -0,0 +1,130 @@
...
+    if (err) {
+        log().error({'err': err}, message);
+        callback(err);
+        return true;
+    } else if (res.statusCode >= 400) {
+        var responseErr = {'code': res.statusCode, 'msg': message, 'body': body};
+        log().error({'err': responseErr}, message);
+        callback(responseErr);
+        return true;
+    }
```
mrvisser: Yea it looks like this error handling pattern will result in the provided callback being executed twice :(

node_modules/oae-zencoder/lib/rest.js
```diff
@@ -0,0 +1,50 @@
...
+ * @Server      tenant
+ * @Method      POST
+ * @Path        /zencoder/encodingNotification
+ * @BodyParam   {notification}    body         Upload metadata from Zencoder
+ * @Return      {void}
+ * @HttpResponse                  200          Encoding completed successfully
+ */
+OAE.tenantRouter.on('post', '/api/zencoder/encodingNotification', function(req, res) {
+    // Tell Zencoder we received the notification or they'll try to resend it
+    res.send(202, {message: 'Notification received'});
```
mrvisser: The `message` key should have single quotes.

node_modules/oae-zencoder/lib/rest.js
```diff
@@ -0,0 +1,50 @@
...
+
+    ZencoderAPI.encodingCompleted(req.body, function(err) {
+        if (err) {
+            log().error('Could not save video previews', err);
```
mrvisser: Is the second parameter contributing anything to the log message? That said, I don't think at this point we have enough information to give a meaningful log message. We only have an opaque request body that we don't necessarily understand inside the rest api. We should probably let ZencoderAPI do the logging since it knows how to extract the content item id, tenant alias and such for a useful log message.

node_modules/oae-zencoder/lib/api.js
```diff
@@ -0,0 +1,137 @@
...
+/**
+ * Save data for the completed preview when encoding has been completed.
+ *
+ * @param  {Object}     notification    The notification data received from Zencoder
+ * @param  {Function}   callback        Standard callback function
+ * @param  {Object}     callback.err    An error that occurred, if any
+ */
+
+var encodingCompleted = module.exports.encodingCompleted = function(notification, callback) {
+    log().info('Received encoding completed notification from Zencoder');
```
mrvisser: Can we wait until we know for what content item or what tenant we have received a notification? This log message on its own seems only useful when doing development and not for debugging a production system when there are are potentially many encoding complete messages coming through.

node_modules/oae-zencoder/lib/api.js
```diff
@@ -0,0 +1,137 @@
...
+    if (job.state === 'finished') {
+        var output = notification.output;
+        var zencoderId = output.id.toString();
+
+        ZencoderDAO.getContentRevisionId(zencoderId, function(err, contentRevision) {
+            if (err) {
+                return callback(err);
+            } else if (!contentRevision || !contentRevision.contentId || !contentRevision.revisionId) {
+                log().warn('No preview job exists for zencoderId %s', zencoderId);
+                return callback('Non-existing Zencoder ID was provided');
```
mrvisser: I think if the content revision didn't exist, then `ZencoderDAO.getContentRevisionId` would have given us an error, so I'm not sure this has to be handled in addition to the standard error case.

node_modules/oae-zencoder/lib/api.js
```diff
@@ -0,0 +1,137 @@
...
+                    return callback(err);
+                }
+
+                return callback();
+            });
+        });
+    }
+
+    else if (job.state === 'failed') {
+        return callback({'err': notification, 'msg': 'Zencoder could not transcode this video for preview'});
```
mrvisser: This should be code/msg style of error

node_modules/oae-zencoder/lib/api.js
```diff
@@ -0,0 +1,137 @@
...
+    }
+
+    else {
+        return callback({'err': notification, 'msg': 'Could not process notification from Zencoder'});
```
mrvisser: Should be code/msg style of error

node_modules/oae-zencoder/lib/rest.js
mrvisser: After looking at the error handling in the API a bit more, if we assume the API will provide sufficient error information on the `err` object itself, then this line can just change to `log().error({'err': err}, 'Failed to save video previews');`

node_modules/oae-zencoder/lib/rest.js
```diff
@@ -0,0 +1,50 @@
...
+            return;
+        }
+
+        log().info('Video preview processing completed successfully');
```
mrvisser: I think we're already logging this in the API.

node_modules/oae-zencoder/lib/util.js
```diff
@@ -0,0 +1,121 @@
...
+
+/**
+ * Get the Zencoder API values that have been configured in the Admin UI.
+ *
+ * @return {Object} The Zencoder configuration for the tenant or the generic config for the host
+ */
+var getConfig = module.exports.getConfig = function(tenantAlias) {
+    var config = ZencoderAPI.getConfiguration();
+
+    if (ZencoderConfig.getValue(tenantAlias, 'zencoder', 'enabled') && ZencoderConfig.getValue(tenantAlias, 'zencoder', 'apiKey')) {
```
mrvisser: Can you first store `isEnabled` and `apiKey` into variables? Then we don't have to refetch the apiKey and it's a bit easier to read.

node_modules/oae-zencoder/lib/util.js
```diff
@@ -0,0 +1,121 @@
...
+        return _.extend(config, {
```
mrvisser: This is going to overwrite the the global system configuration object with potentially tenant-specific values, which will then potentially be shared across tenants.

node_modules/oae-zencoder/lib/util.js
```diff
@@ -0,0 +1,121 @@
...
+            'apiKey': ZencoderConfig.getValue(tenantAlias, 'zencoder', 'apiKey'),
+            'watermark': null,
+            'amazons3-region': ContentConfig.getValue(tenantAlias, 'storage', 'amazons3-region') || config['amazons3-region'],
+            'amazons3-bucket': ContentConfig.getValue(tenantAlias, 'storage', 'amazons3-bucket') || config['amazons3-bucket'],
+            'amazons3-access-key': ContentConfig.getValue(tenantAlias, 'storage', 'amazons3-access-key') || config['amazons3-access-key'],
+            'amazons3-secret-key': ContentConfig.getValue(tenantAlias, 'storage', 'amazons3-secret-key') || config['amazons3-secret-key']
```
mrvisser: Rather than storing the system default amazon configuration, we can use the tenant configuration, while allowing the tenants to override it for their tenant. I believe that's what we did already for mediacore configurations. Is that what is desired here?

node_modules/oae-zencoder/lib/util.js
```diff
@@ -0,0 +1,121 @@
...
+ *
+ * @param  {Object}     config              The default configuration
+ * @param  {String}     tenantAlias         The alias of the tenant whose configuration to use
+ * @param  {String}     path                The signed path to video/audio file
+ * @param  {Function}   callback            Standard callback function
+ * @param  {Error}      callback.err        An error that occurred, if any
+ * @param  {Response}   callback.response   The HTTP Response object
+ * @param  {String}     callback.body       The body of the response
+ */
+var job = module.exports.job = function(tenantAlias, path, contentId, revisionId, isVideo, callback) {
```
mrvisser: `job` isn't all that descriptive of an action, maybe it could be `createJob`

node_modules/oae-zencoder/lib/util.js
```diff
@@ -0,0 +1,121 @@
...
+    var config = getConfig(tenantAlias);
+
+    var tenant = TenantsAPI.getTenant(tenantAlias);
+    var baseUrl = util.format('s3://s3-%s.amazonaws.com/%s/%s', config['amazons3-region'], config['amazons3-bucket'], contentId);
```
mrvisser: Does any URL encoding have to happen here?

node_modules/oae-zencoder/lib/util.js
```diff
@@ -0,0 +1,121 @@
...
+    var extension = isVideo ? 'mp4' : 'mp3';
```
mrvisser: `isVideo` should be in brackets: `(isVideo) ? ...`

node_modules/oae-zencoder/lib/util.js
```diff
@@ -0,0 +1,121 @@
...
+        pass_through: util.format('%s/%s', contentId, filename),
+        private: true,
+        height: 600,
+        output: {
+            notifications: util.format('https://%s/api/zencoder/encodingNotification', tenant.host),
+            base_url: baseUrl,
+            filename: filename,
+            format: extension
+        }
+    };
```
mrvisser: Object keys need quoting

node_modules/oae-preview-processor/lib/processors/link/vimeo.js
sallakarppinen: Because it was wrong; if the video has no thumbnail, we want to ignore it, and to ignore it we need to set `ignore` to `true`.

node_modules/oae-zencoder/lib/util.js
sallakarppinen: That is what's desired, yes. I don't think we did it for mediacore, I think for that if tenants didn't have it set up it just wouldn't work (so no default config).

node_modules/oae-zencoder/lib/util.js
sallakarppinen: I thought _.extend returned a new object? Will check and fix if not.

config.js
sallakarppinen: This raises a more general question about whether we want to use a specific bucket for this or put them in the same S3 bucket as other files. I've been vacillating between the two.

node_modules/oae-zencoder/lib/util.js
```diff
@@ -0,0 +1,121 @@
...
+ */
+var job = module.exports.job = function(tenantAlias, path, contentId, revisionId, isVideo, callback) {
+    var config = getConfig(tenantAlias);
+
+    var tenant = TenantsAPI.getTenant(tenantAlias);
+    var baseUrl = util.format('s3://s3-%s.amazonaws.com/%s/%s', config['amazons3-region'], config['amazons3-bucket'], contentId);
+    var extension = isVideo ? 'mp4' : 'mp3';
+    var filename = util.format('%s.%s', revisionId.replace(/:/g, '-'), extension);
+
+    var options = {
```
mrvisser: Although this options object is probably documented somewhere on zencoders website, can you provide either a direct link or comment docs here indicating what the properties are supposed to represent?

node_modules/oae-zencoder/lib/util.js
```diff
@@ -0,0 +1,121 @@
...
+                            width: '20%'
+                        }
+                    ]
+            });
+        }
+    } else {
+        _.extend(options, {
+                skip_video: true
+        });
+    }
```
mrvisser: All object keys need quotes.

node_modules/oae-zencoder/lib/util.js
```diff
@@ -0,0 +1,121 @@
...
+                            url: config.watermark,
+                            opacity: 0.3,
+                            width: '20%'
+                        }
+                    ]
+            });
+        }
+    } else {
+        _.extend(options, {
+                skip_video: true
```
mrvisser: A single-key object can be done on one line, and there is no space after the `{` or before the `}`.

node_modules/oae-zencoder/lib/util.js
```diff
@@ -0,0 +1,121 @@
...
+        });
+    }
+
+    log().info('Creating Zencoder job');
```
mrvisser: Should contain information that is helpful in debugging issues in production

node_modules/oae-zencoder/lib/util.js
mrvisser: If it helps, this is how we avoid mutating objects w/ `_.extend`: `_.extend({}, config, { ... overrides ... });`

The first parameter is mutated, so this will effective copy the config.

node_modules/oae-zencoder/lib/util.js
```diff
@@ -0,0 +1,121 @@
...
+                    {
+                        label: 'base',
+                        number: 1,
+                        base_url: baseUrl,
+                        height: PreviewConstants.SIZES.IMAGE.LARGE,
+                        width: PreviewConstants.SIZES.IMAGE.LARGE,
+                        filename: 'base'
+                    }
+                ]
+            });
```
mrvisser: This indentation is unconventional to the rest of the codebase.

node_modules/oae-content/lib/internal/util.js
sallakarppinen: AWS is not too clear about it, I'll find out. Currently we don't actually use the aws-sdk elsewhere, but the library we do use is deprecated so we should probably switch to using the aws-sdk more generally as well.

node_modules/oae-zencoder/lib/util.js
mrvisser: It also seems extend isn't necessary here, extending deep nested objects can be confusing w/ underscore as it's not really a merge. We might want to change this to: `options.output.thumbnails = [ ... ]` instead so it's a bit more explicit what's happening.

node_modules/oae-zencoder/lib/util.js
```diff
@@ -0,0 +1,121 @@
...
+        if (config.watermark) {
+            _.extend(options.output, {
+                    watermarks: [
+                        {
+                            url: config.watermark,
+                            opacity: 0.3,
+                            width: '20%'
+                        }
+                    ]
+            });
```
mrvisser: Same here wrt `options.output.watermarks = [ ... ]`

node_modules/oae-zencoder/lib/util.js
mrvisser: This could just change to `options.skip_video = true`

mrvisser:
I've finished a round of review. Looking pretty good so far, there are some concerns mostly around handling of configuration, global object mutation and some error handling.

config.js
mrvisser: We'll probably want it in the same S3 bucket so we can focus our backup efforts on just one S3 bucket.

node_modules/oae-content/lib/internal/util.js
mrvisser: Also, if this involves a remote network call, it should use an asynchronous call rather than synchronous.

node_modules/oae-content/lib/internal/util.js
mrvisser: I agree we need to move to something supported, though if it can't be configured in a multi-threaded environment (using multi-threading loosely here for multiple asynchronous calls), then it's not a real option.

If we aren't using the library elsewhere, it can still cause issues for multiple concurrent calls to get a signed URL, and it will be a really tricky issue to catch if later down the line we use it and the developer and reviewer don't catch this global config gotcha.

node_modules/oae-preview-processor/lib/processors/link/vimeo.js
mrvisser: Ah right, I see that now.

node_modules/oae-zencoder/lib/util.js
mrvisser: @nicolaasmatthijs tenants didn't get mediacore by default?

It's possible that the configuration I'm describing is actually blocked by our oldest PR: https://github.com/oaeproject/Hilary/pull/717

If that's the case, the most reasonable way to accomplish this is probably to fall back to the `config.js` like is happening here. However I think the suggestion to move it to the zencoder config scope is still relevant.

mrvisser: Note that we may also want to provide deployers of the OAE to *not* have a default configuration like this (i.e., videos simply have no processing enabled). Therefore things should fail over rather gracefully if `config.js` is not configured with a default for it.

node_modules/oae-content/lib/internal/util.js
sallakarppinen: Apparently I can configure the bucket on creation instead of using the global AWS configuration, so that solves that one.

node_modules/oae-zencoder/lib/util.js
nicolaasmatthijs: Tenants didn't get MediaCore by default. For any tenant that didn't have MediaCore, it wouldn't attempt to generate a preview for video files.

In this particular case, the intention for *Unity is to give all tenants a default ZenCoder API configuration (with limited use) and allow them to override this with their tenant specific ZenCoder configuration. However, the ESUP deployment will not have a default ZenCoder API configuration and will require individual tenants to set this.

Therefore, I think we just want to have a ZenCoder config that can be set at the global admin level and can be overridden at the tenant level.

node_modules/oae-content/lib/internal/util.js
```diff
@@ -68,12 +70,14 @@ var getStorageBackend = module.exports.getStorageBackend = function(ctx, uri) {
...
- * @param  {Number}     [offset]    The minimum time in seconds for which the generated picture URLs will be valid. Default: 1 week
+ * @param  {Context}    ctx             Standard context object containing the current user and the current tenant
+ * @param  {Content}    content         The content object that needs to be augmented with signatures and download urls
+ * @param  {Number}     [duration]      The approximate time in seconds for which the generated picture URLs will be valid. The larger this value is, the more effective browser caching is on the download which is good for thumbnail images. Default: 1 week
+ * @param  {Number}     [offset]        The minimum time in seconds for which the generated picture URLs will be valid. Default: 1 week
+ * @param  {Function}   callback        Standard callback function
+ * @param  {Object}     callback.err    An error that occurred, if any
  */
-var augmentContent = module.exports.augmentContent = function(ctx, content, duration, offset) {
+var augmentContent = module.exports.augmentContent = function(ctx, content, duration, offset, callback) {
```
mrvisser: Looks like the changes here and to the docs can be reverted

node_modules/oae-content/lib/internal/util.js
```diff
@@ -104,7 +108,45 @@ var augmentContent = module.exports.augmentContent = function(ctx, content, dura
...
+
+/**
+ * Generate a signed url that can be used to show video previews in the browser for a video file stored in AWS S3
+ *
+ * @param  {Context}    ctx               Current request context
+ * @param  {String}     content           The content object that needs a signed S3 video URL
+ * @param  {Function}   callback          Standard callback function
+ * @param  {Object}     callback.err      An error that occurred, if any
+ */
+var getSignedVideoUrl = module.exports.getSignedVideoUrl = function(ctx, content, callback) {
```
mrvisser: Since the intention of this method is not to get something, but to add it to the content object, maybe we could name it `applySignedVideoUrl`. Alternatively, if this method just took in a ctx and video key and returned the video URL (the caller would be responsible for what to do with it -- assert it in a test, apply it to the content object, etc...) then the function name could stay the same.

node_modules/oae-content/lib/internal/util.js
```diff
@@ -104,7 +108,45 @@ var augmentContent = module.exports.augmentContent = function(ctx, content, dura
...
+    if (!content.previews.videoKey) {
+        return callback();
     }
+
+    var key = content.previews.videoKey;
```
mrvisser: Rather than access `content.previews.videoKey` twice, we could first do `var key = content.previews.videoKey;` and then check `if (!key) {...` afterward.

node_modules/oae-preview-processor/lib/api.js
```diff
@@ -507,18 +507,18 @@ var _generatePreviews = function(ctx, callback) {
...
+                    ctx.setStatus(status, callback);
                 } else {
                     ctx.savePreviews(callback);
                 }
             });
         } else {
             log().info({'contentId': ctx.contentId, 'content': ctx.content, 'revision': ctx.revision}, 'Ignoring as this type of content is not supported for now');
-            ctx.setStatus('ignored', callback);
+            ctx.setStatus(ContentConstants.previews.IGNORED, callback);
         }
```
mrvisser: A bit of a nitpick, but this recommendation is in the spirit of avoiding creating asynchronous branches inside if statements:

We have 2 "escape" procedures here where we don't want to just log / promote something then break right away: the error case and the "ignored" case. Then we have our happy path which is to run the generate previews function. We typically want to do the escape routes first: stop processing immediately and promote the error condition. Then our happy path can live outside the escape conditions (i.e., not within the if statement) in the main body of the function. So it might look more like this:

```javascript
if (err) {
  return callback(err);
} else if (!processor) {
  log().info({'contentId': ctx.contentId, 'content': ctx.content, 'revision': ctx.revision}, 'Ignoring as this type of content is not supported for now');
  return ctx.setStatus(ContentConstants.previews.IGNORED, callback);
}

processor.generatePreviews(ctx, ctx.content, function(err, status) {
  if (err) {
    return ctx.setStatus('error', callback);
  } else if (ignored) {
    return ctx.setStatus(ContentConstants.previews.IGNORED, callback);
  } else if (status && _.contains(ContentConstants.previews, status)) {
    return ctx.setStatus(status, callback);
  }

  return ctx.savePreviews(callback);
});
```

node_modules/oae-preview-processor/lib/api.js
```diff
@@ -507,18 +507,18 @@ var _generatePreviews = function(ctx, callback) {
         if (err) {
             return callback(err);
         } else if (processor) {
-            processor.generatePreviews(ctx, ctx.content, function(err, ignored) {
+            processor.generatePreviews(ctx, ctx.content, function(err, status) {
                 if (err) {
-                    ctx.setStatus('error', callback);
-                } else if (ignored) {
-                    ctx.setStatus('ignored', callback);
+                    ctx.setStatus(ContentConstants.previews.ERROR, callback);
```
mrvisser: I think this should be `ContentConstants.previews.IGNORED`

node_modules/oae-preview-processor/lib/api.js
```diff
@@ -507,18 +507,18 @@ var _generatePreviews = function(ctx, callback) {
...
+                } else if (status && _.contains(ContentConstants.previews, status)) {
+                    ctx.setStatus(status, callback);
```
mrvisser: What does this case concretely represent? Is there a way to check for it more directly, or if not add some code comments to explain?

node_modules/oae-zencoder/config/config.js
```diff
@@ -0,0 +1,31 @@
...
+        'description': 'Configuration for Zencoder video transcoding',
+        'elements': {
+            'apiKey': new Fields.Text('API Key', 'The Zencoder API key for your account', '', {'suppress': true, 'globalAdminOnly': true}),
+            'amazons3-bucket': new Fields.Text('Amazon S3 Bucket', 'The Amazon S3 bucket for output videos', '', {'suppress': true, 'globalAdminOnly': true}),
+            'amazons3-access-key': new Fields.Text('Amazon Access Key', 'Your Amazon Access key', '<access-key>', {'suppress': true, 'globalAdminOnly': true}),
+            'amazons3-secret-key': new Fields.Text('Amazon Secret Key', 'Your Amazon Secret key', '<secret-key>', {'suppress': true, 'globalAdminOnly': true}),
+            'amazons3-region': new Fields.Text('Amazon S3 Region', 'The region for your S3 bucket', 'eu-west-1', {'suppress': true, 'globalAdminOnly': true})
+        }
+    }
+};
```
mrvisser: Does this configuration allow for the tenant admin to see it? I suspect `globalAdminOnly` stops them from being able to set something. Rather than test it out on the UI, this is something that would be good to write some unit tests to verify and we can keep them in place afterward.

node_modules/oae-zencoder/lib/api.js
```diff
@@ -0,0 +1,122 @@
...
+var ZencoderDAO = require('./internal/dao');
+
+/**
+ * Save data for the completed preview when encoding has been completed.
+ *
+ * @param  {Object}     notification    The notification data received from Zencoder
+ * @param  {Function}   callback        Standard callback function
+ * @param  {Object}     callback.err    An error that occurred, if any
+ */
+
```
mrvisser: unneeded space

node_modules/oae-zencoder/lib/api.js
```diff
@@ -0,0 +1,122 @@
...
+var encodingCompleted = module.exports.encodingCompleted = function(notification, callback) {
+    var job = notification.job;
+    var output = notification.output;
+    var zencoderId = output.id.toString();
+
+    if (job.state !== 'finished') {
+        return callback({'code': 500,'msg': 'Zencoder could not transcode video with zencoderId ' + zencoderId});
```
mrvisser: `util.format` with placeholders should be used instead of concatenation.

node_modules/oae-zencoder/lib/api.js
```diff
@@ -0,0 +1,122 @@
...
+    } else if (!output.url || !job.pass_through || !output.thumbnails) {
+        return callback({'code': 500, 'msg': 'Received malformed notification from Zencoder'});
```
mrvisser: If something does ever go wrong, it might help to know which keys specifically we expected that weren't available on the request object.

node_modules/oae-zencoder/lib/api.js
```diff
@@ -0,0 +1,122 @@
...
+    }
+
+    ZencoderDAO.getContentRevisionId(zencoderId, function(err, contentRevision) {
+        if (err) {
+            return callback(err);
+        }
+
+        return _createAndSavePreviews(contentRevision, output, job['pass_through'], callback);
```
mrvisser: This should be `job.pass_through` for consistency.

node_modules/oae-zencoder/lib/api.js
```diff
@@ -0,0 +1,122 @@
...
+ * @param  {String}     key               The path to the video file on Amazon S3
+ * @param  {Function}   callback          Standard callback function
+ * @param  {Object}     callback.err      An error that occurred, if any
+ * @api private
+ */
+var _createAndSavePreviews = function(contentRevision, output, key, callback) {
+    var contentId = contentRevision.contentId;
+    var ctx = new PreviewContext(PreviewProcessorAPI.getConfiguration(), contentId, contentRevision.revisionId);
+
+    log().info('Received encoding completed notification from Zencoder for contentId %s', contentId);
```
mrvisser: Something like an object id is a little more handy to have in the "data" portion of the log:

`log().info({'contentId': contentId}, 'Received encoding completed notification from Zencoder);`

Then the bunyan JSON log parsing utility can allow you to do things like filter that data etc...

node_modules/oae-zencoder/lib/api.js
```diff
@@ -0,0 +1,122 @@
...
+        }
+        PreviewUtil.generatePreviewsFromImage(ctx, path, {'removeInput': true}, function(err) {
+            if (err) {
+                return callback(err);
+            }
+            log().info('Saved preview and thumbnails for video content item %s', ctx.content.contentId);
+            return ctx.savePreviews();
+        });
+    });
+};
```
mrvisser: Files need to end w/ a new line. FWIW, there is an `.editorconfig`, that describes small spacing things like this and editors like Sublime and Atom have plugins to manage it for you.

node_modules/oae-zencoder/lib/init.js
```diff
@@ -13,28 +13,31 @@
...
+ * Create the schema for the Zencoder preview processor
  *
  * @param  {Function}   callback        Standard callback function
  * @param  {Object}     callback.err    An error that occurred, if any
  * @api private
  */
 var _ensureSchema = function(callback) {
     Cassandra.createColumnFamilies({
-        'MediaCoreContentRevisionIdMap': 'CREATE TABLE "MediaCoreContentRevisionIdMap" ("mediaCoreId" text PRIMARY KEY, "contentId" text, "revisionId" text)'
+        'ZencoderToContentIdMap': 'CREATE TABLE "ZencoderToContentIdMap" ("zencoderId" text PRIMARY KEY, "contentId" text, "revisionId" text)'
```
mrvisser: What should we do with current data surrounding media core? We probably have content items whose preview metadata references media core, how does the system react when someone tries to view the preview?

I'm assuming we'll want to reprocess them, does the zencoder stuff just overwrite the mediacore metadata?

node_modules/oae-zencoder/lib/processor.js
```diff
@@ -0,0 +1,85 @@
...
+ * permissions and limitations under the License.
+ */
+
+var _ = require('underscore');
+var fs = require('fs');
+var request = require('request');
+var util = require('util');
+
+var ContentUtil = require('oae-content/lib/internal/util');
+var ContentConstants = require('oae-content/lib/constants').ContentConstants;
```
mrvisser: Alphabetical ordering is off here

node_modules/oae-zencoder/lib/processor.js
```diff
@@ -0,0 +1,85 @@
...
+            log().error({'err': err, 'msg': 'Error occurred when posting video to Zencoder'});
+            return callback(err);
+        } else if (res.statusCode >= 400) {
+            var responseErr = {'code': res.statusCode, 'body': body};
+            log().error({'err': responseErr, 'msg': util.format('Received status code %s when posting video to Zencoder', res.statusCode)});
+            return callback(responseErr);
+        } else if (!body.outputs) {
+            var bodyError = {'err': body, 'msg': 'Received malformed response body from Zencoder'};
+            log().error(bodyError);
+            return callback(bodyError);
```
mrvisser: We can probably give a more detailed message about the lack of outputs field

node_modules/oae-preview-processor/lib/api.js
sallakarppinen: This is the error condition - if there was an error in the preview processing, we should set status to 'error', correct?

node_modules/oae-zencoder/config/config.js
mrvisser: When we do make it viewable by a tenant, should we keep the ability to make it explicitly "enabled" / "disabled" in configuration? I think that's missing from what we had in the mediacore config.

node_modules/oae-zencoder/lib/util.js
```diff
@@ -0,0 +1,129 @@
...
+    var apiKey = ZencoderConfig.getValue(tenantAlias, 'zencoder', 'apiKey');
+
+    if (apiKey) {
+        return _.extend({}, config, {
+            'apiKey': apiKey,
+            'watermark': null,
+            'amazons3-region': ZencoderConfig.getValue(tenantAlias, 'zencoder', 'amazons3-region') || config['amazons3-region'],
+            'amazons3-bucket': ZencoderConfig.getValue(tenantAlias, 'zencoder', 'amazons3-bucket') || config['amazons3-bucket'],
+            'amazons3-access-key': ZencoderConfig.getValue(tenantAlias, 'zencoder', 'amazons3-access-key') || config['amazons3-access-key'],
+            'amazons3-secret-key': ZencoderConfig.getValue(tenantAlias, 'zencoder', 'amazons3-secret-key') || config['amazons3-secret-key']
```
mrvisser: Here we're kind of doing a merge of tenant configuration values with system configuration values. Did you consider strictly using one or the other based on if the tenant is enabled? I'm not sure which approach is better, but here's a scenario:

a) A tenant puts in their API key, region, bucket, access key but not the secret key (so, it's their public key and our secret key)
b) They invoke some requests that ultimately cause request failures against their account
c) Is there potentially some way for them to see am audit log of what secret key was used against their amazon account?

Further more, adding default values "under the hood" (i.e., they can't see the values we're applying here, not even in the admin ui) could make configuration issues a bit trickier to solve for them.

node_modules/oae-zencoder/lib/init.js
sallakarppinen: I think we should be able to use the admin interface to reprocess specific mime-types? As far as I've understood, it'll overwrite the mediacore metadata, We'll probably need a script to remove the mediacore tables and configuration as well when all the code has been taken out?

node_modules/oae-zencoder/config/config.js
sallakarppinen: I think I've gotten confused with the config settings - I was going for an option where a) normal users can not see the configuration, b) tenant admins can not see the default values, c) tenant admins can change the values for their own and can see those values after the change.

node_modules/oae-zencoder/lib/processor.js
```diff
@@ -0,0 +1,85 @@
...
+    var revision = ctx.revision;
+
+    if (revision.previews && revision.previews.videoKey) {
+        return callback();
+    }
+
+    log().info('Generating previews for video file: %s', content.id);
+    var tenant = content.tenant;
+
+    var downloadPath = ContentUtil.getSignedDownloadUrl(new Context(tenant), revision.uri, 6000, 6000);
```
mrvisser: Does Zencoder have any documentation on when it would be able to get to download this file? Do we have any guarantee it would get there before this expires?

node_modules/oae-zencoder/lib/util.js
```diff
@@ -0,0 +1,129 @@
...
+
+    var tenant = TenantsAPI.getTenant(tenantAlias);
+    var baseUrl = util.format('s3://s3-%s.amazonaws.com/%s/%s', config['amazons3-region'], config['amazons3-bucket'], encodeURIComponent(contentId));
+    var extension = (isVideo) ? 'mp4' : 'mp3';
+    var filename = util.format('%s.%s', revisionId.replace(/:/g, '-'), extension);
+
+    // @See https://app.zencoder.com/docs/api/encoding for details of Zencoder's encoding settings
+    var options = {
+        'input': util.format('https://%s%s', tenant.host, path),
+        'pass_through': util.format('%s/%s', contentId, filename), // We store the S3 bucket filepath so it can later be used to create a signed URL
```
mrvisser: We can put the comment above the object property. Same-line inline comments is not something we've embraced stylistically.

mrvisser: Are we sure we need to use pass_through? At the time that we get the response, we get the zencoder id, which lets us get the content and revision id which is all we need to derive the object key.

node_modules/oae-zencoder/lib/util.js
```diff
@@ -0,0 +1,129 @@
...
+        ];
+        if (config.watermark) {
+            options.output.watermarks = [
+                {
+                    'url': config.watermark,
+                    'opacity': 0.3,
+                    'width': '20%'
+                }
+            ];
+        }
```
mrvisser: Do we have a use-case for the watermark? If not, maybe we should omit any logic and configuration associated to it for now or else we'll need to write tests with and without it to ensure we don't have any landmines.

node_modules/oae-zencoder/lib/util.js
```diff
@@ -0,0 +1,129 @@
...
+    } else {
+        options.skip_video = true;
+    }
+
+    log().info('Creating Zencoder job for revision %s', revisionId);
```
mrvisser: revisionId can go into a log data property

mrvisser:
I've finished another round of review. I think general approach looks good and we can start getting some unit tests going for it.

node_modules/oae-preview-processor/lib/api.js
mrvisser: I misread the diff! Sorry about that.

node_modules/oae-zencoder/config/config.js
mrvisser: I don't think we are able to set up the configuration such that the global admin can set values that the tenant administrators cannot see, but they can override it. There is a really old PR open to do this but never had sufficient attention.

To avoid that from blocking this feature, I think we'll need to have the "global admin" values in the `config.js` (as you have right now), and set up this configuration such that:

* Tenant admins can override the values
* Regular users cannot see the values

... which I think is just `suppress: true`.

The global admin then is just not supposed to set the global admin overrides for this features, as it is intended to fall back to the `config.js`.

node_modules/oae-zencoder/lib/init.js
mrvisser: Maybe. I think it's just a matter of testing to ensure that those files get reprocessed appropriately.

node_modules/oae-zencoder/lib/processor.js
sallakarppinen: No, they don't, but I think it's unlikely that they would not. On testing the whole process seems to have mostly been a few minutes at tops. I can extend the time though if you think it's better.

node_modules/oae-zencoder/lib/util.js
sallakarppinen: We'd need the extension as well, which depends on whether it's a video or audio file. It's easier to pass it than re-create it.

node_modules/oae-zencoder/lib/util.js
sallakarppinen: It came up in discussions with Jim when first starting with the module and exists as an encouragement for tenants to configure their own Zencoder account or accept a Unity logo on their videos when viewing them in Unity.

sallakarppinen:
Made more changes based on the latest round of comments.

node_modules/oae-zencoder/lib/util.js
mrvisser: Don't we also derive the extension based on information we already have locally after the callback? I don't think we would be recreating it would we, as we'd just create it the one time when the receive the "finished" result callback?

sallakarppinen: `pass_through` is what the filepath ends up being on S3, with contentId as the 'folder' and revisionId plus the extension as the filename. We need a value for these in util.js, since that's what we're using to tell Zencoder where to put the file. In api.js we could get the filepath from the url the file ended up on by getting the substring after the third /, or we could recreate a string with the contentId, revisionId with : replaced with -, and an extension depending on `notification.output.format`, or we could use `pass_through`. I don't see a benefit to the first two options.

mrvisser: > that's what we're using to tell Zencoder where to put the file

Oh, Zencoder is pushing the file directly to S3 for us -- I missed that. I'm guessing the credential access key and secret key need to be configured on their end? Because we don't appear to be sending any access credentials here.

sallakarppinen: They have credentials to add files to the bucket set up with a bucket policy (as described [here](https://app.zencoder.com/docs/guides/getting-started/working-with-s3)).

sallakarppinen:
After thinking about it some more, I'm pretty certain we can't let tenants set up their own S3 buckets for videos... Simply because we won't have any way of telling whether the preview for any one file is in the tenant bucket or the global bucket (where all the previews created before the tenant set up their own bucket will have ended up). I think I'm going to take the option to configure the bucket out. @mrvisser, any thoughts?

nicolaasmatthijs:
@sallakarppinen : Can you also add some Mixpanel tracking for this (e.g., length of uploaded video) so we can easily track how much video transcoding time every tenant has used?

sallakarppinen:
@nicolaasmatthijs: Yes. Do we want the video duration in milliseconds or seconds (Zencoder returns it in ms, but seconds would probably be more readable)?

nicolaasmatthijs:
@sallakarppinen : I'd be fine with either. Do they charge per second of video? If so, tracking seconds might be easiest.

sallakarppinen:
@nicolaasmatthijs: Per minute, actually, with partial minutes handled in 10 second increments rounded up.

nicolaasmatthijs:
Okay. It might make sense to track the time in seconds, but also track the cost based on how Zencoder calculates the cost.

sallakarppinen:
The cost varies based on subscription type and is not guaranteed to stay the same, so I would argue against hardcoding it.

mrvisser:
@sallakarppinen I had a look at the tests and its coverage on the code, and I think there are a few areas that needs to see some test coverage. I've attached the coverage report (it's generated using `grunt test-coverage`), you can see the report by opening `target/lcov-report/index.html` in the archive: [test-coverage.tar.gz](https://github.com/oaeproject/Hilary/files/432437/test-coverage.tar.gz)

Some examples of coverage that it seems to lack:

* Trying to generate video for a tenant that has it disabled (i.e., the getConfig is never invoked with a disabled tenant)
* The watermark functionality never gets covered (missing coverage in `util.js / createJob`)
* The skip_video functionality never gets covered (missing coverage in `util.js / createJob`)
* Verifying handling of missing pass_through property on response (missing coverage in `encodingCompleted`)
* No coverage in the `rest.js` api

... and so on.

node_modules/oae-zencoder/tests/test-zencoder-api.js
```diff
@@ -0,0 +1,196 @@
...
+    describe('#encodingCompleted', function() {
+        it('verify a valid notification is processed correctly', function(callback) {
+            TestsUtil.generateTestUsers(globalAdminOnTenantRestContext, 1, function(err, users) {
+                assert.ok(!err);
+                var user = _.values(users)[0];
+                var restCtx = user.restContext;
+                // Create the content item that we'll test with
+                RestAPI.Content.createFile(restCtx, 'test video', null, 'private', _getVideoStream, null, null, null, function(err, content) {
+                    assert.ok(!err);
+                    ZencoderDAO.saveContentRevisionId('10', content.id, content.latestRevisionId, function(err) {
```
mrvisser: Rather than doing this stuff manually, we should allow the preview-processing system to complete this for us -- a lot of this is because we run our unit tests as integration tests.

This wasn't happening for you before probably because previews are disabled by default in the unit tests. If you enable the preview processor, then you'll find that you can make use of the full preview processing subsystem. Please have a look at the tests for things like the link processor and pdf processor etc... and how they upload files to the API and then wait for them to be processed, then assert the results.

mrvisser: Here is an example of some of our preview tests: https://github.com/oaeproject/Hilary/blob/master/node_modules/oae-preview-processor/tests/test-previews.js#L356

You can see how we bail on them if previews aren't enabled in the config.js. There are utilities shown in there that wait for messages to make it through rabbitmq etc...

I think the easiest way to test may be to use nock to mock the expected HTTP requests to zencoder while making use of the full preview processing lifecycle, and use the `RestAPI` to stage our callbacks.

Additionally, after the callbacks, it will be important to verify the preview status and data on the revisions to ensure it persists properly, I think those kinds of assertions are currently be overlooked here as well.

mrvisser:
Assigning back to @sallakarppinen for more extensive "integration-like" preview lifecycle testing.

node_modules/oae-zencoder/tests/test-zencoder-api.js
sallakarppinen: Should these kinds of tests go under the preview-processor then? The ones I've written were basically following the same pattern as the ones we currently have for mediacore [here](https://github.com/oaeproject/Hilary/blob/master/node_modules/oae-mediacore/tests/test-mediacore-processor.js#L143), with the exception of using nock to catch the requests instead of spinning up a server to respond in lieu of Zencoder. It seems like tests for the actual preview processing process should go into that module instead of replicating the setup here.

mrvisser: To re-use utilities that are being used in the preview processor tests, it would probably be easier to put the tests in there. I think that would be fine.

sallakarppinen: Cool - I'll keep these more 'unit-testy' tests here and extend them to cover the cases you mentioned above, and add a more integrationary one in the preview-processor module.

PR #1296, opened 2016-08-19
Backend for adding tenant logo via admin site

sallakarppinen:
Allow for tenant logo to be uploaded directly into OAE, rather than requiring an scp into the production environment.


node_modules/oae-ui/lib/api.js
```diff
@@ -811,6 +813,37 @@ var _replaceOptimizedPaths = function(skinVariables) {
...
+ */
+var uploadLogoFile = module.exports.uploadLogoFile = function(ctx, file, tenantAlias, callback) {
+    // Need to require this here to avoid circular dependency issues
+    var ContentAPI = require('oae-content/lib/api');
+    tenantAlias = tenantAlias || ctx.tenant().alias;
+    var tenant = ctx.tenant();
+    if (!ctx.user() || !ctx.user().isAdmin(tenantAlias)) {
+        return callback({'code': 401, 'msg': 'Only administrators can upload new logos for tenants'});
+    }
+    ContentAPI.createFile(ctx, util.format('%s logo', tenantAlias), util.format('Logo displayed on %s landing page', tenantAlias), AuthzConstants.visibility.PRIVATE, file, {}, [], function(err, content) {
```
simong: Can we also validate the uploaded file is an `image/*` mime type?

nicolaasmatthijs:
@mrvisser @simong : Rather than creating a custom endpoint for uploading a logo, would it be better to just enable the generic content upload endpoint for the global admin tenant?

node_modules/oae-ui/lib/api.js
```diff
@@ -811,6 +813,37 @@ var _replaceOptimizedPaths = function(skinVariables) {
...
+        if (err) {
+            return callback({'code': 500, 'err': err});
```
simong: I think the ContentAPI will already return an OAE error object (`{'code': .., 'err': err}`) so you should just be able to do `return callback(err);` here.

node_modules/oae-ui/lib/api.js
```diff
@@ -811,6 +813,37 @@ var _replaceOptimizedPaths = function(skinVariables) {
...
+        }
+        ContentAPI.getRevision(ctx, content.id, content.latestRevisionId, function(err, revision) {
+            if (err) {
+                return callback({'code': 500, 'err': err});
```
simong: Same thing here

node_modules/oae-ui/tests/test-ui.js
```diff
@@ -561,6 +561,37 @@ describe('UI', function() {
...
+                fileStream = fs.createReadStream(__dirname + '/data/oae-logo.png');
+                // Assert that a tenant admin can change the logo for a tenant
+                RestAPI.UI.uploadLogo(camAdminRestContext, fileStream, tenantAlias, function(err, data) {
+                    assert.ok(!err);
+                    assert.ok(data);
+                    assert.ok(data.url.indexOf('signed') !== -1);
+
+                    fileStream = fs.createReadStream(__dirname + '/data/oae-logo.png');
+                    // Assert that a random user can not change the logo for a tenant
+                    RestAPI.UI.uploadLogo(anonymousCamRestContext, fileStream, tenantAlias, function(err, data) {
```
simong: Can you try with an authenticated user too, just to cover our bases?

node_modules/oae-ui/lib/api.js
simong: Maybe we should change this too:
```javascript
ContentUtil.getStorageBackend(ctx).store(ctx.tenant().alias, file, options, function(err, uri) {
```

That way the logo won't end up in the search index or the admin's content library.

simong:
@nicolaasmatthijs Possibly, but it's probably safer and cleaner to have an endpoint that only takes in a file and stores it with the configured storage backend. If we re-use the content create endpoint the logo will be searchable, have a content profile etc. The UI will also have to ensure it's sending it with a private content visibility.

node_modules/oae-ui/lib/api.js
sallakarppinen: We are validating the mime type in the front end, but I'll add a backend check as well.

nicolaasmatthijs:
@simong : That makes sense. However, I was thinking there might be some advantages to being able to rely on the library/search/etc. functionality. First of all, it would allow us to upload other skin-related items (e.g., background image) and it would allow an admin to pick from the files they've already uploaded (e.g., switch back to regular logo after having uploaded a special occasion logo).

We would have to make sure that global admin tenant content is excluded on regular tenant searches though.

What do you think?

simong:
This PR allows doesn't limit you to just the tenant logo, you could use it to upload background images as well as the UI is still responsible for setting the actual config values (ie: set the URL for a tenant logo, set the URL for a background image)

That said, if there's a requirement to be able to pick from your own content library, than yes, it would probably make more sense to enable those content endpoints on the global admin.

There are a few todos with this:
 - Do we expose an endpoint that generates a signature for a public content item that never expires? Otherwise, we'd have to use something like `/api/content/:content_id/download/:revision_id` which will run a few DB queries for each image that gets displayed by the skin.
 - We'll have to add the necessary exclusions when dealing with `admin` content in the search handlers.
 - We should probably also add some tests that files shouldn't be removed from the storage backend when deleting the content items, otherwise we'd end up with a broken UI.
 - What happens when you upload a new revision for a content item? Does it update the skin or is the skin's image bound to a specific revision?

sallakarppinen:
Pushed some changes based on comments here.

@simong: 

>   Do we expose an endpoint that generates a signature for a public content item that never expires? Otherwise, we'd have to use something like /api/content/:content_id/download/:revision_id which will run a few DB queries for each image that gets displayed by the skin.

I *think* we kind of do, depending on the backend, [here](https://github.com/oaeproject/Hilary/blob/master/node_modules/oae-content/lib/rest.js#L958).

>  We'll have to add the necessary exclusions when dealing with admin content in the search handlers.

If the files are created with private visibility, wouldn't that cover this case?

>   We should probably also add some tests that files shouldn't be removed from the storage backend when deleting the content items, otherwise we'd end up with a broken UI.

Agreed.

>  What happens when you upload a new revision for a content item? Does it update the skin or is the skin's image bound to a specific revision?

Previously when using ContentAPI.createFile and getRevision, is was tied to a revision. I'm not totally sure yet about how ContentUtil.getStorageBackend.store works, so will need to have another look at that.

nicolaasmatthijs:
>> We'll have to add the necessary exclusions when dealing with admin content in the search handlers.
> If the files are created with private visibility, wouldn't that cover this case?

The files will at least need to be publicly accessible so they can be loaded from the tenant. Or is the idea that the file will be private but the access token will be included in the configured logo URL?

sallakarppinen:
@nicolaasmatthijs: Yes, the idea is that the file is private but has a non-expiring signed URL that can be used as the logo URL.

node_modules/oae-ui/tests/test-ui.js
```diff
@@ -561,6 +561,56 @@ describe('UI', function() {
                 });
             });
         });
+
+         /**
+         * Test that verifies that a new logo can be uploaded for a tenant and a signed URL is returned
+         */
+        it('verify logo can be uploaded for a tenant', function(callback) {
+            var tenantAlias = global.oaeTests.tenants.cam.alias;
+            var fileStream = fs.createReadStream(__dirname + '/data/oae-logo.png');
```
brecke: A quick suggestion: Since `__dirname + '/data/oae-logo.png'` is repeated several times in this test, why not create a local variable for the file path and re-use it?

node_modules/oae-ui/lib/api.js
```diff
@@ -811,6 +813,38 @@ var _replaceOptimizedPaths = function(skinVariables) {
...
+        return callback({'code': 401, 'msg': 'Only administrators can upload new logos for tenants'});
+    }
+    var extension = file.name.split('.').pop();
+    if (!extension.match(/(gif|jpe?g|png)$/i)) {
+        return callback({'code': 500, 'msg': 'File has an invalid mime type'});
+    }
+
+    var options = {
+        'prefix': util.format('logos/%s', tenantAlias)
+    }
```
sallakarppinen: I'd leave it for now since it's not something that's been changed in this PR (and I'm not sure about 2 either).

PR #1313, opened 2017-02-02
Preventing modification of the name and mail for CAS and Shib

dooremont:
replace #1193

---
This change is [Reviewable](https://reviewable.io/reviews/oaeproject/hilary/1313)

### DIFF
--- a/node_modules/oae-authentication/config/strategies.js
+++ b/node_modules/oae-authentication/config/strategies.js
@@ -58,7 +58,8 @@ module.exports = {
         'description': 'Allow Shibboleth authentication for tenant',
         'elements': {
             'enabled': new Fields.Bool('Shibboleth Authentication Enabled', 'Allow Shibboleth authentication for tenant', false),
-            'name': new Fields.Text('Name', 'A name that users will recognize as their identity provider', ''),
+            'disableProfile': new Fields.Bool('Preventing modification of the name and mail', 'Preventing modification of the name and mail', true),
+	    'name': new Fields.Text('Name', 'A name that users will recognize as their identity provider', ''),
             'idpEntityID': new Fields.Text('Identity Provider entity ID', 'The entity ID of the IdP', '', {'suppress': true}),
             'externalIdAttributes': new Fields.Text('External ID Attribute', 'The attribute that uniquely identifies the user. This should be a prioritised space seperated list', 'eppn persistent-id targeted-id', {'suppress': true}),
             'mapDisplayName': new Fields.Text('Display name', 'The attibute(s) that should be used to construct the displayname. This should be a prioritised space seperated list. e.g., `displayname cn`', 'displayname cn', {'suppress': true}),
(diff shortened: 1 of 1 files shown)


node_modules/oae-authentication/config/strategies.js
```diff
@@ -58,7 +58,8 @@ module.exports = {
         'description': 'Allow Shibboleth authentication for tenant',
         'elements': {
             'enabled': new Fields.Bool('Shibboleth Authentication Enabled', 'Allow Shibboleth authentication for tenant', false),
-            'name': new Fields.Text('Name', 'A name that users will recognize as their identity provider', ''),
+            'disableProfile': new Fields.Bool('Preventing modification of the name and mail', 'Preventing modification of the name and mail', true),
+	    'name': new Fields.Text('Name', 'A name that users will recognize as their identity provider', ''),
```
brecke: Bad indent

node_modules/oae-authentication/config/strategies.js
```diff
@@ -71,6 +72,7 @@ module.exports = {
         'description': 'Allow CAS authentication for tenant',
         'elements': {
             'enabled': new Fields.Bool('CAS Authentication Enabled', 'Allow CAS authentication for tenant', false),
+	    'disableProfile': new Fields.Bool('Preventing modification of the name and mail', 'Preventing modification of the name and mail', true),
```
brecke: Bad indent

node_modules/oae-authentication/config/strategies.js
brecke: Can we also change `Preventing ...` to `Prevent ...`? It feels more natural.

node_modules/oae-authentication/config/strategies.js
brecke: Same comment on `Preventing` becoming `Prevent` as above

brecke:
@dooremont have you considered the possibility of a malicious user changing the HTML code on a page to simulate an `input` field instead of a `label` so he can try to submit data? 

The fix on [4120](https://github.com/oaeproject/3akai-ux/pull/4120) "hides" the form, but we still need to make sure the backend doesn't change the data if CAS or Shibboleth are being used, in case an unexpected `POST` (form submission) comes in.

brecke:
Hi,

So, what I tried to explain above is that we need to make sure that the server behaves correctly if the `POST` comes in. The logic you implemented on the client:

```
if disabled:
  then...
if not:
  then...
```

isn't present on the server.

The fact that the HTML has been changed in order to make the right things appear to the user doesn't mean that it can't be simulated via javascript, for instance (hence the picture above, which proves that one can still submit the form and cheat the server). We need to be protecting **both the client and the server**.

davidoae:
Note that the client can never be trusted as the code is easily modifiable
by the user.
Whatever security considerations are in the client are irrelevant from the
perspective of the server, the server has to assume a malicious client.

Sorry if I'm butting in unnecessarily :)

D

dooremont:
## how to get element config in api.user.js

I search the method to get config element disableProfile _(Hilary/node_modules/oae-authentication/config/strategies.js )_
  ```
shibboleth': {
     
'disableProfile': new Fields.Bool('Preventing modification of the name and mail', 'Preventing modification of the name and mail', true),
```

in api.user.js for the if

thanks a lot for your help

**Comment by brecke** [comment]
Hi @dooremont!

I'm not sure I understood your question. What is it you're looking for?

**Comment by dooremont** [comment]
ok i search the method to get a config element for exemple "disableProfile" in file Hilary/node_modules/oae-authentication/config/strategies.js 

i want implement the test process
```
if disabled:
  then...
if not:
  then...

```
in api.user.js file

**Comment by brecke** [comment]
hi!

I think you're looking for this kind of [statement](https://github.com/oaeproject/3akai-ux/blob/master/node_modules/oae-admin/adminlogin/js/adminlogin.js#L77) which applied to your specific case would become something like this:

```
var foo = oae.api.config.getValue('oae-authentication', 'strategies', 'disableProfile');
```

Was that it?

dooremont:
exactly thanks a lot
best

brecke:
@ClaireLozano will you take over this as well? I mean once all the ongoing PRs are merged, of course.

brecke:
Start by rebasing, as usual, and go from there. I've assigned this to you.

PR #1318, opened 2017-02-08
fix #1317

### LINKED ISSUE #1317: Make sure there are no &apos; in e-mails
We know that some e-mail clients such as Outlook don't render the message with `html5`. As such, we need to make sure we're sending a `html4` compliant entity instead of `&apos;`

brecke:
Now replacing &apos; for html4 compliant entity

### DIFF
--- a/node_modules/oae-email/lib/api.js
+++ b/node_modules/oae-email/lib/api.js
@@ -383,6 +383,11 @@ var sendEmail = module.exports.sendEmail = function(templateModule, templateId, 
 
     if (renderedTemplates['html']) {
         emailInfo.html = renderedTemplates['html'];
+
+        // We need to escape the &apos; entity because some e-mail clients
+        // don't (yet) support html5 rendering, such as Outlook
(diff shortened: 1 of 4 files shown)


node_modules/oae-email/lib/api.js
```diff
@@ -383,6 +383,11 @@ var sendEmail = module.exports.sendEmail = function(templateModule, templateId,
 
     if (renderedTemplates['html']) {
         emailInfo.html = renderedTemplates['html'];
+
+        // We need to escape the &apos; entity because some e-mail clients
+        // don't (yet) support html5 rendering, such as Outlook
+        // Tip from http://stackoverflow.com/questions/419718/html-apostrophe
+        emailInfo.html = emailInfo.html.replace(/&apos;/g, '&#39;');
```
sallakarppinen: Won't this just replace the first instance of `&apos;` but not the following ones?

sallakarppinen:
Could you write a test for this?

PR #1329, opened 2017-02-24
Added nodemon support

brecke:
Now the backend restarts itself every time a file is saved
Install nodemon with `npm install --save-dev nodemon`
Run the app with `nodemon app.js | bunyan`


nodemon.json
```diff
@@ -0,0 +1,37 @@
+{
+  "verbose": true,
+  "watch": [
+    "node_modules/oae-activity",
```
sallakarppinen: Can we try using a regex (something like oae-*) instead? Otherwise we will need to update this file every time we change the modules (for example, the Jitsi PR adds a module and the Zencoder PR adds one and removes one).

nodemon.json
```diff
@@ -0,0 +1,37 @@
...
+    "node_modules/oae-tests",
+    "node_modules/oae-tincanapi",
+    "node_modules/oae-ui",
+    "node_modules/oae-uservoice",
+    "node_modules/oae-util",
+    "node_modules/oae-version"
+  ],
+  "ignoreRoot": [".git"],
+  "ext": "js json"
+}
```
sallakarppinen: Newline.

package.json
```diff
@@ -140,11 +140,12 @@
     "istanbul": "latest",
     "mocha": "latest",
     "nock": "2.17.0",
+    "nodemon": "^1.11.0",
```
sallakarppinen: We could possibly use `latest` here - as far as I understand, since `nodemon` doesn't really interfere with our code, it shouldn't lead to incompatibilities.

nodemon.json
brecke: I tried :) I'll give it another go later

package.json
brecke: I agree

brecke: Woops! It turns out there's a good reason for having the caret range, which is the possibility of updating to `v2.x` and the config file no longer works (`v2.x` might not be backwards compatible). So I say we keep it as is because it's safer this way.

sallakarppinen: Cool beans.

Issue #1340, opened 2017-03-15
Create method of fetching all users for a tenancy


sallakarppinen:
Will probably require an extra table in Cass.

davidoae:
Really? Duplicating data in my mind seems wrong.
Does the api not already have an endpoint for this?

Of course I'm just interfering without good knowledge.

D

sallakarppinen:
The API does not have an endpoint for this, and there's not really any way to get the data without duplicating it considering the NoSQL nature of Cassandra... Just querying all users by `tenancyAlias` is not possible.

brecke:
What if we make `tenantAlias` a primary key? I don't even know if that's possible, just thinking out loud.

PR #1342, opened 2017-03-17
Fetching all users for a tenant

sallakarppinen:
Fixes issue #1340.

### DIFF
--- a/node_modules/oae-principals/lib/rest.user.js
+++ b/node_modules/oae-principals/lib/rest.user.js
@@ -784,3 +784,30 @@ OAE.tenantRouter.on('get', '/api/user/:userId/groups/recent', function(req, res)
         return res.send(200, {'results': results});
     });
 });
+
+/**
+ * @REST getUsersForTenant
+ *
+ * Get all users for a given tenant
+ *
+ * @Server      admin,tenant
+ * @Method      GET
+ * @Path        /tenant/{tenantAlias}/users
+ * @PathParam   {string}        tenantAlias                 The id of the user to get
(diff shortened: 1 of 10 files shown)


node_modules/oae-principals/lib/rest.user.js
```diff
@@ -784,3 +784,30 @@ OAE.tenantRouter.on('get', '/api/user/:userId/groups/recent', function(req, res)
...
 });
+
+/**
+ * @REST getUsersForTenant
+ *
+ * Get all users for a given tenant
+ *
+ * @Server      admin,tenant
+ * @Method      GET
+ * @Path        /tenant/{tenantAlias}/users
```
brecke: Let's make it `tenants` as plurals are the [best and common practice](http://www.restapitutorial.com/lessons/restfulresourcenaming.html).

node_modules/oae-principals/lib/rest.user.js
```diff
@@ -784,3 +784,30 @@ OAE.tenantRouter.on('get', '/api/user/:userId/groups/recent', function(req, res)
...
+    PrincipalsAPI.getAllUsersForTenant(req.ctx, req.params.tenantAlias, function(err, users) {
+        if (err) {
+            return res.send(err.code, err.msg);
+        }
+
+        return res.send(200, users);
+    });
+};
+
+OAE.globalAdminRouter.on('get', '/api/tenant/:tenantAlias/users', _getUsersForTenant);
```
brecke: Same as above

PR #1364, opened 2017-04-13
Fixes #1363

### LINKED ISSUE #1363: Fix the disable users by tenancy migration
Doesn't run due to a dependency error

brecke:
Added check to avoid dependency error
Added a log error with the uncaughtException stack trace
Minor fixes to the migration

Test it by running `node etc/migration/disable_users_from_tenancy/1.disable-users-by-tenancy.js -t someTenant | bunyan`

### DIFF
--- a/etc/migration/disable_users_from_tenancy/1.disable-users-by-tenancy.js
+++ b/etc/migration/disable_users_from_tenancy/1.disable-users-by-tenancy.js
@@ -84,11 +84,11 @@ OAE.init(config, function (err) {
 
   DisableUsersMigration.doMigration(null, tenantAlias, true, function(err, users) {
     if (err) {
-      log().warn("Migration not completed successfully.");
+      log().warn('Migration not completed successfully.');
       process.exit(err.code);
     }
 
-    log().info("Finished migration for " + users.length + " users.");
+    log().info('Finished migration for ' + users.length + ' users.');
 
     // Nothing left to do, exiting.
     process.exit(0);
--- a/etc/migration/disable_users_from_tenancy/1.disable-users-by-tenancy.js
+++ b/etc/migration/disable_users_from_tenancy/1.disable-users-by-tenancy.js
@@ -1,3 +1,5 @@
+#!/usr/bin/env node
+
 /*!
  * Copyright 2017 Apereo Foundation (AF) Licensed under the
  * Educational Community License, Version 2.0 (the "License"); you may
@@ -22,6 +24,7 @@ var optimist = require('optimist');
 var path = require('path');
 var util = require('util');
 
+var AuthzConstants = require('oae-authz/lib/constants').AuthzConstants;
 var Context = require('oae-context').Context;
 var log = require('oae-logger').logger('oae-script-main');
 var OAE = require('oae-util/lib/oae');
@@ -47,7 +50,7 @@ var argv = optimist
 
 if (argv.help) {
   optimist.showHelp();
-  return process.exit(1);
+  return;
 }
 
 // Get the config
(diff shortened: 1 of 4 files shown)


brecke:
Refactored it a bit

node_modules/oae-principals/lib/model.user.js
```diff
@@ -43,6 +43,7 @@ var TenantsAPI = require('oae-tenants');
  * @param  {Boolean}    [opts.isTenantAdmin]            Whether or not the user is a tenant admin
  */
 module.exports.User = function(tenantAlias, id, displayName, email, opts) {
+    var TenantsAPI = TenantsAPI || require('oae-tenants');
```
sallakarppinen: Can we take this out if we are generating a context in the migration script?

etc/migration/disable_users_from_tenancy/lib/disable-users-by-tenancy.js
```diff
@@ -29,6 +37,9 @@ var PrincipalsAPI = require('oae-principals');
  * @param  {Function}   callback        Standard callback function
  */
 var doMigration = function (ctx, tenantAlias, disabled, callback) {
+
+    ctx = ctx || _createNewContext();
```
sallakarppinen: This does seem a bit unsafe but should be OK for a migration.

node_modules/oae-principals/lib/model.user.js
brecke: oops. I forgot to remove this

etc/migration/disable_users_from_tenancy/lib/disable-users-by-tenancy.js
brecke: why unsafe?

sallakarppinen: Well, `_createNewContext` generates a new global admin context which would be not-good anywhere except in a migration script, OK here though.

etc/migration/disable_users_from_tenancy/1.disable-users-by-tenancy.js
```diff
@@ -83,19 +81,14 @@ OAE.init(config, function (err) {
...
-  var globalAdmin = new User(globalTenant.alias, util.format('u:%s:admin', globalTenant.alias), 'Global Administrator', null, {
-    'visibility': AuthzConstants.visibility.PRIVATE,
-    'isGlobalAdmin': true
-  });
 
-  DisableUsersMigration.doMigration(new Context(globalTenant, globalAdmin), tenantAlias, true, function(err, users) {
+  DisableUsersMigration.doMigration(null, tenantAlias, true, function(err, users) {
     if (err) {
-      log().warn('Migration not completed successfully.');
+      log().warn("Migration not completed successfully.");
```
sallakarppinen: Should use single quotes.

etc/migration/disable_users_from_tenancy/1.disable-users-by-tenancy.js
```diff
@@ -83,19 +81,14 @@ OAE.init(config, function (err) {
...
       process.exit(err.code);
     }
 
-    log().info('Finished migration for ' + users.length + ' users.');
+    log().info("Finished migration for " + users.length + " users.");
```
sallakarppinen: Here as well, single quotes for all.

node_modules/oae-util/lib/oae.js
```diff
@@ -61,6 +61,7 @@ module.exports.init = function(config, callback) {
 
     // Set up application-level error handler
     process.on('uncaughtException', function(err) {
+        log().error(err.stack);
```
sallakarppinen: These logging statements could be swapped around so that the stack trace shows up below the message.

Issue #1355, opened 2017-04-10
Links should not preview hosts that resolve to private IPs


davidoae:
The links preview can be used to discover information about the OAE private network setup.
As deployments can use any private address group all private IPs and hosts that resolve to private IPs should not be previewed. It *may* also be reasonable to make this partially configurable so that a deployment can mask it's own private network even if the IPs used are public ones.

10.0.0.0/8
192.168.0.0/16
172.16.0.0/12
100.64.0.0/10
192.0.2.0/24
198.51.100.0/24
203.0.113.0/24
198.18.0.0/15
127.0.0.0/8
fd00::/8

davidoae:
I think we should try to have a fix for this on the next release.

sallakarppinen:
I started looking into this while in London but got sidetracked with the LTI stuff - I'll get back on it tomorrow assuming nothing urgent pops up.

sallakarppinen:
Closed by #1372

PR #1372, opened 2017-05-19
Disclude internal IPs from link previews

sallakarppinen:
Fixes issue #1355


davidoae:
Thanks SK, looks good afaict.

PR #1374, opened 2017-05-24
Stop launch from failing if VersionAPI call fails

sallakarppinen:


node_modules/oae-lti/lib/api.js
```diff
@@ -48,7 +48,12 @@ var getLtiTool = module.exports.getLtiTool = function(ctx, toolId, groupId, call
             }
             VersionAPI.getVersion(function(err, version) {
                 if (err) {
-                    return callback(err);
+                    log().warn('Failed to fetch OAE version')
```
brecke: semicolon missing

node_modules/oae-lti/lib/api.js
```diff
@@ -48,7 +48,12 @@ var getLtiTool = module.exports.getLtiTool = function(ctx, toolId, groupId, call
...
+                    version = {
+                        hilary: {
+                            version: ''
+                        }
+                    }
```
brecke: semicolon missing

Issue #1396, opened 2017-06-14
Add a test for password reset feature


brecke:
To make sure that the new password follows the same rules as signup (has been implemented, but there is no test for that)

sallakarppinen:
This was added in #1399

PR #1399, opened 2017-06-20
Added tests and changed paths for password reset

sallakarppinen:


sallakarppinen:
Fixes issue #1396

sallakarppinen:
I'll put them to a single `it` statement - not too keen on using a `before` since I want to do some assertions in the common part as well.

brecke:
No need for another round of review if the test is passing 👍

Issue #1422, opened 2017-08-09
Users being created with no `tenantAlias`


sallakarppinen:
There have been users created with almost no fields; the affected user records only contain `principalId`, `notificationsUnread` and `lastModified` fields. This causes an error when trying to access the user from Hilary, when attempting to fetch their tenancy: 
```ERROR: oae/29038 on app0:
    TypeError: Cannot call method 'compact' of null
     at new module.exports.User (/opt/oae/node_modules/oae-principals/lib/model.user.js:56:26)
     at _hashToUser (/opt/oae/node_modules/oae-principals/lib/internal/dao.js:818:16)
     at _getUserFromRow (/opt/oae/node_modules/oae-principals/lib/internal/dao.js:796:12)
     at _getPrincipalFromRow (/opt/oae/node_modules/oae-principals/lib/internal/dao.js:763:76)
     at Function._.map._.collect (/opt/oae/node_modules/underscore/underscore.js:172:24)
     at _.(anonymous function) [as map] (/opt/oae/node_modules/underscore/underscore.js:1496:34)
     at /opt/oae/node_modules/oae-principals/lib/internal/dao.js:217:14
     at /opt/oae/node_modules/oae-util/lib/cassandra.js:783:16
     at onReturn (/opt/oae/node_modules/helenus/lib/connection.js:490:7)
     at onReturn (/opt/oae/node_modules/helenus/lib/connection.js:422:7)
```
We should investigate where these records are being created from and add additional checks to stop this from happening again.

PR #1423, opened 2017-08-10
Adds check to not create `unreadNotifications` for deleted users

sallakarppinen:
So, the issue was that after records had been deleted from `Principals`, they were still members of groups and had activities created for them when something happened in those groups. This led to partial records being created in the `Principals` table, which then caused errors when Hilary tried to access those records (as they were missing required fields like `tenantAlias`). I've added a check to stop users from being collected if the records are invalid, which I think should stop the errors from happening; the other thing we need to do is to make sure we delete `AuthzRoles` records as well when deleting users manually. This should no longer be necessary when the task to permanently delete users is completed, but we should also keep an eye on the issue when reviewing that task.


sallakarppinen:
Fixes #1422

PR #1430, opened 2017-08-22
Changed how we configure env

sallakarppinen:
Unfinished - just wanted to check that the approach is what @brecke had in mind when first raising this. I've added `dotenv` and I imagine this would lead to us incrementally moving to .env files in puppet-hilary as well for environmentally dependant stuff, so would also like comments from @davidoae.


sallakarppinen:
Righto, as dotenv doesn't support nested variables (which we do quite rely on), it seems that it's not the tool for us. Instead, I propose the following: we add another config file to the repository called 'development.js' with the dev settings. As this file is loaded after config.js, it should override values in that. We can then slowly move environment dependent vars to similarly named files in puppet-hilary (qa.js for qa etc) instead of using puppet & hiera to rewrite config.js. Does this make sense to people outside my head?

brecke:
Yes, it does make sense 🕺

sallakarppinen:
Cool - this should now work. There's a related PR in oaeproject/puppet-hilary#242.

PR #1435, opened 2017-09-05
Issue 1434

brecke:
screen shot 2017-09-05 at 16 23 01

### DIFF
--- a/repl.js
+++ b/repl.js
@@ -1,3 +1,22 @@
+/*!
+ * Copyright 2017 Apereo Foundation (AF) Licensed under the
+ * Educational Community License, Version 2.0 (the "License"); you may
+ * not use this file except in compliance with the License. You may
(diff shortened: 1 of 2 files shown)


repl.js
```diff
@@ -0,0 +1,55 @@
...
+ * or implied. See the License for the specific language governing
+ * permissions and limitations under the License.
+ */
+
+/*
+ * Usage: `node repl.js` and then `PrincipalsAPI.createUser(...);` e.g.
+ */
+
+let repl = require("repl").start({}),
+    promisify = require("repl-promised").promisify;
```
sallakarppinen: I'd prefer these as 
```javascript
const repl = require('repl').start({});
const promisify = require('repl-promised').promisify;
```

Additionally, shouldn't repl-promised be added to package.json?

brecke: Fixed!

PR #1437, opened 2017-09-12
Updated submodule oae-rest

brecke:


greysteil:
@brecke - would you be up for automating submodule updates? I just added that functionality to [Dependabot](https://dependabot.com) and would love some beta testers!

brecke:
@greysteil how would that help me exactly? This issue was the result of not pushing the parent when a submodule was updated.

For this type of scenario, what I need is a notification mechanism telling me whenever the parent submodule pointers are not up to date, instead of finding it the hard way. We use the "recursive push" mechanism but sometimes this fails e.g. if one developer opens up a PR in one of the submodules and not on the parent.

greysteil:
So, with Dependabot, it will check every day whether the parent is up-to-date. If it isn't it will put in a pull request to the parent with the updated submodule reference.

sallakarppinen:
> I think that's what you need here, and would have saved you having to create this PR manually?

Yeah, it probably would've. What would being a beta tester look like here?

greysteil:
If you click through at [Dependabot](https://dependabot.com) you'll be able to install it to this account. Then you need to click "Add Project", select Hilary, and choose "Git Submodules" as the language. You'll be able to give feedback on anything Dependabot could do better by @-mentioning it in the comments.

The tool is totally free for open-source and always will be, but if you help us out with beta testing I'll make sure it's free for you for private repos, too :)

sallakarppinen:
Cool beans. I've installed it for Hilary and we'll see how it goes. 👍 Thanks!

brecke:
So far so good, I guess that it's really useful to us!

PR #1442, opened 2017-09-22
Set a different fetchSize value for cassandra queries

### LINKED ISSUE #1441: Bypass the fetchSize: 5000 limitation when using cassandra driver
We need to overcome [this](https://github.com/datastax/nodejs-driver#paging).

brecke:
Fixes #1441

### DIFF
--- a/node_modules/oae-config/lib/api.js
+++ b/node_modules/oae-config/lib/api.js
@@ -399,7 +399,7 @@ var _cacheAllTenantConfigs = function(callback) {
  * @param  {Object}     callback.tenantConfigs  An object keyed by tenant alias, whose value is a key value pair of `configKey->value` for each stored configuration value for the tenant
  */
 var _getAllPersistentTenantConfigs = function(callback) {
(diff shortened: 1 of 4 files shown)


sallakarppinen:
Is this a permanent solution? If not, I think we should spend a couple of days looking into a more robust option instead of adding technical debt.

sallakarppinen:
Thanks. I think everything that ends up being added to the elasticsearch indices needs to use this method as well, since that eats whole tables.

brecke:
Let's see if this works first and then I'll take care of the remaining caching methods 👍

brecke:
@sallakarppinen I've applied the same querying technique to `Principals` and `Config`. I wasn't able to spot any other caching methods that would "eat whole tables", can you find any. 

All tests pass locally and on circleCI except for the flickr thing.

sallakarppinen:
Did you check all the search index methods? The Flickr issue is on their end, but no reply on when it might be fixed yet.

brecke:
For reference: [Flickr API user group](https://www.flickr.com/groups/api/discuss/72157686417438430/)

Issue #1257, opened 2015-10-27
Make test log level configurable


mrvisser:
Currently, our tests hardcode the log level to trace:

https://github.com/oaeproject/Hilary/blob/master/node_modules/oae-tests/lib/util.js#L983-L986

However we should make this configurable. To avoid having to do string replace the config file, it might be easiest to just use an `OAE_TEST_*` environment variable. We already have environment variables such as:

```
OAE_TEST_INTEGRATION
OAE_TEST_DROP_KEYSPACE_BEFORE
OAE_BOOTSTRAP_LOG_LEVEL
OAE_BOOTSTRAP_LOG_FILE
```

Maybe it's fine to have:

```
OAE_TEST_LOG_LEVEL
OAE_TEST_LOG_PATH
```

Which default to `trace` / `./tests.log` if unspecified.

PR #1452, opened 2017-10-04
#1257 - Retrieve test log level from environment

nywleswoey:
This is with reference to issue #1257 to make the test log level
configurable. If OAE_TEST_LOG_LEVEL or OAE_TEST_LOG_PATH are
not present in the environment, they would be defaulted to 'trace' and
'./test.log' respectively.

### DIFF
--- a/local.js
+++ b/local.js
@@ -30,7 +30,7 @@ var config = module.exports.config = {};
  */
 config.test = {
     'timeout': 60000,
-    'level': process.env.OAE_TEST_LOG_LEVEL || 'info',
-    'path': process.env.OAE_TEST_LOG_PATH || './tests.log'
+    'level': 'info',
+    'path': './tests.log'
 };
 
--- a/local.js
+++ b/local.js
@@ -25,7 +25,12 @@ var config = module.exports.config = {};
  * Configuration namespace for the OAE tests.
  *
  * @param  {String}    timeout            The mocha timeout that should be used
+ * @param  {String}    level              The log level that should be used for testing
+ * @param  {String}    path               The log path that should be used for testing
  */
 config.test = {
-    'timeout': 60000
+    'timeout': 60000,
+    'level': process.env.OAE_TEST_LOG_LEVEL || 'info',
+    'path': process.env.OAE_TEST_LOG_PATH || './tests.log'
 };
+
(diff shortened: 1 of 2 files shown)


sallakarppinen:
Hello @nywleswoey and thanks for taking this on! I have two comments: firstly, could we have the default level as `info` - `trace` and even `debug` tend to generate too much noise to be very useful. Secondly, we've changed the way we do environment-specific configuration so it's a bit different now... Have a look [here](https://github.com/oaeproject/Hilary/blob/master/node_modules/oae-tests/lib/util.js#L967). To explain what's going on there, we set the environment using `process.env.NODE_ENV` to something like `local`, then have the settings for that environment in a file called `local.js` in the root of the app [here](https://github.com/oaeproject/Hilary/blob/master/local.js). Does that make sense?

nywleswoey:
Hello @sallakarppinen. Thanks for the reply.
Sure, I could change the log level to info. I set it as 'trace' because the actual issue description stated so. Regarding the local.js, please see my latest commit is done correctly. Thanks again!

sallakarppinen:
Yeah, we should've updated the task... Sorry about that.

local.js
```diff
@@ -25,7 +25,12 @@ var config = module.exports.config = {};
  * Configuration namespace for the OAE tests.
  *
  * @param  {String}    timeout            The mocha timeout that should be used
+ * @param  {String}    level              The log level that should be used for testing
+ * @param  {String}    path               The log path that should be used for testing
  */
 config.test = {
-    'timeout': 60000
+    'timeout': 60000,
+    'level': process.env.OAE_TEST_LOG_LEVEL || 'info',
```
sallakarppinen: These can just be `'info'` and `./tests.log` - they won't be set in `process.env` anymore.

node_modules/oae-tests/lib/util.js
```diff
@@ -983,8 +983,8 @@ var createInitialTestConfig = module.exports.createInitialTestConfig = function(
 
     // log everything (except mocha output) to tests.log
     config.log.streams = [{
-        'level': 'info',
-        'path': './tests.log'
+        'level': config.test.level,
```
sallakarppinen: Could you set these to `config.test.level || 'info'` and `config.test.path || `'./tests.log'`? The default values should be here instead of in the env file, so that they work in environments other than local, too.

nywleswoey:
Thanks Salla.

PR #1454, opened 2017-10-12
Changed reconnection and load balancing for Cassandra

sallakarppinen:
What it says on the tin; this *may* help with the test issues. There's a couple of things here I'd like to check... Firstly I'm not sure if we want to have `consistency: 'two'` or if we should set it to some other value, or even not set it at all in the client and allow instances to specify it for themselves in cassandra.yaml... Secondly, I'm not sure if we want to keep the `readTimeout` value here - see comment in the [client options doc](http://docs.datastax.com/en/developer/nodejs-driver/3.3/api/type.ClientOptions/).

### DIFF
--- a/node_modules/oae-util/lib/cassandra.js
+++ b/node_modules/oae-util/lib/cassandra.js
@@ -96,10 +96,9 @@ var _createNewClient = function (hosts, keyspace) {
         keyspace: keyspace,
         protocolOptions: { maxVersion: 3 },
         socketOptions: {
-            readTimeout: CONFIG.timeout,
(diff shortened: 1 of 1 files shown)


brecke:
Right, let's have a go at it like this and go from there.

Regarding consistency, I have tried with `queryOptions: { consistency: types.consistencies.quorum }` and got the same results :(

Let's also try removing the timeout then and see where that leads us.

sallakarppinen:
Cool, I've changed consistency and taken out the read timeout.

PR #1502, opened 2018-01-09
Fixes #1501

### LINKED ISSUE #1501: CircleCI doesn't pass for any branches other than master
The `oae-version` test fails all the time if we're testing some branch other than master, because of this line:

```
- run: git submodule foreach -q --recursive 'branch="$(git config -f $toplevel/.gitmodules submodule.$name.branch)"; [ -z "$branch" ] && git checkout master || git checkout $branch'
```

Which means that we have to change the `oae-version` so that it expects `3akai-ux` to be a submodule and not a branch.

brecke:
Removed unnecessary line from CCI config
Slight change in test
Refactored a bit

Hopefully CCI will pass tests starting now ^_^

### DIFF
--- a/node_modules/oae-version/lib/api.js
+++ b/node_modules/oae-version/lib/api.js
@@ -13,27 +13,28 @@
  * permissions and limitations under the License.
  */
 
-var _ = require('underscore');
-var fs = require('fs');
-var gift = require('gift');
-var path = require('path');
-var util = require('util');
+const _ = require('underscore');
+const fs = require('fs');
+const gift = require('gift');
+const path = require('path');
+const util = require('util');
 
-var IO = require('oae-util/lib/io');
-var log = require('oae-logger').logger('oae-version');
+const IO = require('oae-util/lib/io');
+const log = require('oae-logger').logger('oae-version');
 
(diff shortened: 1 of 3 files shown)


node_modules/oae-version/lib/api.js
```diff
@@ -79,20 +80,48 @@ var getVersion = module.exports.getVersion = function(callback) {
...
-    _getVersionInfo(_uiPath, callback);
+
+    let submodulePointer = "";
+    let lastCheckedOutCommit = "";
+
+    gift.init(hilaryDirectory, (err, repo)  => {
+        repo.tree().contents((err, children) => {
+
+            // find the 3akai-ux submodule within the tree
+            const submodulePointer = _.find(children, (eachChild) => {
```
sallakarppinen: You checked that this still works when 3akai is *not* a submodule? (For example when installing from package?)

brecke: It won't work in any other situation other than the git submodule, because we're creating a reference CI/CD procedure which uses git-based installs. If the need for package installs arise in the future, we might have to re-think this approach, but hopefully that won't be necessary. Merging because all checks passed 🍾

PR #1528, opened 2018-02-20
Fixes #1527

### LINKED ISSUE #1527: Update settings for pm2 and git based deployment
[PM2](http://pm2.keymetrics.io/docs/usage/quick-start/) acts as both a nodemon replacement in development mode as well as a more robust way of running the app in production

brecke:
Changed frontend path in config.js


config.js
```diff
@@ -28,7 +28,7 @@ var config = module.exports.config = {};
  * @param  {String}    path            The path to the UI static assets
  */
 config.ui = {
-    'path': '../3akai-ux'
+    'path': './3akai-ux'
```
sallakarppinen: Are we sure we want to change the UI path for everyone?

brecke: Well, git submodules have been in place for some time now, and the default development environment makes use of this config as well. 

The production deployment will also use submodules (as it is git-based) so it also needs this setting. At this stage we're changing it in every possible environment and scenario, so why not change it for good?

sallakarppinen: Mainly I'm worried about it breaking other people's deployments if they're grown to rely on the default config. But fair point, proceed at will.

brecke: Well, @dooremont told me he checks config every time they upgrade it, which I think is a very sensible thing to do. Plus, their deployment is going to change soon too.

But you raise an important point and I think we should place this in the release notes next chance we have. 👍

Issue #1407, opened 2017-06-21
Documents in a folder should be shared with its group by default


brecke:
From [uservoice](https://oaeproject.uservoice.com/forums/249738-open-academic-environment/suggestions/8858758-documents-in-a-folder-should-be-shared-with-its-gr):

> When you create a document inside of a folder, the document correctly receives the visibility of the folder as the default visibility. But, the documents in the folder are not shared with the group by default. It should have the same sharing settings as that of the folder.

## Testing

Note: The ~Inherit folder invitations?~ represents the fact that I have decided not to inherit invitations (made on the folder) to the content created / uploaded on that folder. The permissions are inherited (editor, viewer, manager) but the invitations are not.

- [x] Collabdoc
  - [x] Users only
  - [x] Users and invitations
  - [ ] ~Inherit folder invitations?~
- [x] Collabsheet
  - [x] Users only
  - [x] Users and invitations
  - [ ] ~Inherit folder invitations?~
- [x] Single Link
  - [x] Users only
  - [x] Users and invitations
  - [ ] ~Inherit folder invitations?~
- [x] Multiple Links
  - [x] Users only
  - [x] Users and invitations
  - [ ] ~Inherit folder invitations?~
- [x] Single Upload
  - [x] Users only
  - [x] Users and invitations
  - [ ] ~Inherit folder invitations?~
- [x] Multiple uploads
  - [x] Users only
  - [x] Users and invitations
  - [ ] ~Inherit folder invitations?~

ClaireLozano:
Hi ! 

I made a pull request here : https://github.com/oaeproject/Hilary/pull/1548 

It works but I broke the unit tests of the module "oae-content" ... Can you review the changes before I repair the unit test.

ClaireLozano:
Hmm ... almost all (in oae-content) :
> 17 passing (33s)
 > 170 failing

I didn't ran the other tests

brecke:
Uhh.. that's goint to take some work then. But before that, I think we need to redesign the solution to this.

The current behaviour is content being shared with the group as read-only, correct? All the group members can see the document, but just as a "viewer", not as an editor. The problem lies on the fact that they are never made editors even if they have "manage" access on the folder. So, with that in mind, here's what we need to to:

1. We need a way to make it clear to the user that he can inherit the writing permissions when creating a document, and make it default behaviour. I'd suggest something like this:

[...]

PR #1548, opened 2018-03-06
Documents in a folder should be shared with its group by default

ClaireLozano:
Issue : https://github.com/oaeproject/Hilary/issues/1407

---
This change is [Reviewable](https://reviewable.io/reviews/oaeproject/hilary/1548)

### DIFF
--- a/node_modules/oae-content/lib/api.js
+++ b/node_modules/oae-content/lib/api.js
@@ -539,25 +539,67 @@ var _createContent = function(ctx, contentId, revisionId, resourceSubType, displ
             return callback(err);
         }
 
-        // The current user always becomes a manager
-        roles[ctx.user().id] = AuthzConstants.role.MANAGER;
-
-        // Create the resource
-        var createFn = _.partial(ContentDAO.Content.createContent, contentId, revisionId, ctx.user().id, resourceSubType, displayName, description, visibility, otherValues, revisionData);
-        ResourceActions.create(ctx, roles, createFn, function(err, content, revision, memberChangeInfo) {
+        // If the content is on a folder, add those members to the content created 
+        _getFolderMembers(ctx, folders[0].id, roles, function(err, roles) {
             if (err) {
                 return callback(err);
             }
 
-            // Add the content item to the specified folders, if any
-            _addContentItemToFolders(ctx, content, folders, function(err) {
+            // The current user always becomes a manager
+            roles[ctx.user().id] = AuthzConstants.role.MANAGER;
+
+            // Create the resource
+            var createFn = _.partial(ContentDAO.Content.createContent, contentId, revisionId, ctx.user().id, resourceSubType, displayName, description, visibility, otherValues, revisionData);
+            ResourceActions.create(ctx, roles, createFn, function(err, content, revision, memberChangeInfo) {
                 if (err) {
-                    log().warn({'err': err, 'contentId': content.id, 'folders': folders}, 'Could not add a content item to a folder');
+                    return callback(err);
                 }
 
-                return callback(null, content, revision, memberChangeInfo);
+                // Add the content item to the specified folders, if any
+                _addContentItemToFolders(ctx, content, folders, function(err) {
+                    if (err) {
+                        log().warn({'err': err, 'contentId': content.id, 'folders': folders}, 'Could not add a content item to a folder');
+                    }
+
+                    return callback(null, content, revision, memberChangeInfo);
+                });
             });
         });
+    });
+};
+
+/**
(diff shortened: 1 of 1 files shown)


node_modules/oae-content/lib/api.js
```diff
@@ -539,28 +539,77 @@ var _createContent = function(ctx, contentId, revisionId, resourceSubType, displ
...
         }
 
-        // The current user always becomes a manager
-        roles[ctx.user().id] = AuthzConstants.role.MANAGER;
-
-        // Create the resource
-        var createFn = _.partial(ContentDAO.Content.createContent, contentId, revisionId, ctx.user().id, resourceSubType, displayName, description, visibility, otherValues, revisionData);
-        ResourceActions.create(ctx, roles, createFn, function(err, content, revision, memberChangeInfo) {
+        // If the content is on a folder, add those members to the content created 
+        _getFolderMembers(ctx, folders, roles, resourceSubType, function(err, roles) {
```
brecke: aha! here's the problem. These are the roles that are currently valid, but this is not what we want, is it? we want the permissions that have been defined by the user in the "shared with" form. Those are the ones we want.

A good strategy would be to pick the current ones and then filter them using the form. That way we would have the folder roles, e.g.

```
{ 'u:up:SkmX6OL9M': 'manager', 'u:uc:HyDW6O8qf': 'editor' }
```

and then compare it to what comes through the form. For instance, if one user has been removed from the input field inside the modal box, we take it out and it becomes 

```
{ 'u:up:SkmX6OL9M': 'manager' }
```

brecke: The POST that is sent when one creates the document goes like this (in my local scenario):

```
resourceSubType:collabdoc
displayName:My private fifth document
description:
visibility:private
managers:u:up:SkmX6OL9M
managers:u:up:SkmX6OL9M
folders:f:up:rkqGJYUqG
```

So you see, that's the baseline we should be using: `{ 'u:up:SkmX6OL9M': 'manager' }`. The current folder permissions doesn't matter.

On another test, where I didn't remove any users from the "Shared with" field, this is what I got:

```
resourceSubType:collabdoc
displayName:My public document
description:
visibility:private
managers:u:up:SkmX6OL9M
managers:u:uc:HyDW6O8qf 👈 
managers:u:up:SkmX6OL9M
folders:f:up:rkqGJYUqG
```

...which confirms my approach: this is what we need `roles` to contain. 👍

ClaireLozano: I finally removed all my code (no code, no mistake haha) because I get all members on 3akai-ux and there is no need to add code in Hilary but there is something that still bothers me... On a collabdoc, if a user is added on the "setpermissions" modal, they will be added as manager. 

-> So if there is 30 viewers on a folder, when a collabdoc is created inside the folder, this collabdoc will have 30 managers (+ the creator of the resource). 

For me, a viewer of a folder should be editor(or viewer) but not manager ... I don't know, that's just my point of view.

(Just for information, viewers and managers of folder will become viewers of a link created inside the folder.)

brecke:
:lgtm:

---

Review status: all files reviewed at latest revision, 1 unresolved discussion, some commit checks failed.

---



*Comments from [Reviewable](https://beta.reviewable.io/reviews/oaeproject/hilary/1548#-:-L9qV8U59pHusOLVtLlp:bnfp4nl)*

brecke:
Review status: all files reviewed at latest revision, 1 unresolved discussion, some commit checks failed.

---

*[node_modules/oae-content/lib/api.js, line 543 at r1](https://beta.reviewable.io/reviews/oaeproject/hilary/1548#-L8fvq3P02bJgAC_7aXA-r1-543:-L9qX91g2jrxfwbofpbN:b-7did8g) ([raw file](https://github.com/oaeproject/hilary/blob/807851dc48ebd899e76e270d0afde88afb36ec26/node_modules/oae-content/lib/api.js#L543)):*
<details><summary><i>Previously, brecke (Miguel Laginha) wrote…</i></summary><blockquote>

> For me, a viewer of a folder should be editor(or viewer) but not manager ... I don't know, that's just my point of view.

I agree. We need to make a folder viewer into a collabdoc editor.
</blockquote></details>

This is the only comment from this review.. but I'm not sure this is done or whether this was just an idea, you tell me

---


*Comments from [Reviewable](https://beta.reviewable.io/reviews/oaeproject/hilary/1548)*

brecke:
Reviewed 1 of 1 files at r2.
Review status: all files reviewed at latest revision, all discussions resolved, some commit checks failed.

---



*Comments from [Reviewable](https://beta.reviewable.io/reviews/oaeproject/hilary/1548)*

PR #1625, opened 2018-11-22
refactor(docker): replaced pdf2html binary with a docker

### LINKED ISSUE #1624: Dockerize pdf2htmlEX dependency
Follow-up to #1341 so that onboarding and setup is simpler and we can avoid system dependencies for Hilary

brecke:
Added new container to docker-compose
Changed things a bit for webshot binary to be explicitly set

Fixed #1624

---
This change is [Reviewable](https://reviewable.io/reviews/oaeproject/hilary/1625)

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -2,7 +2,9 @@ version: 2
 jobs:
   test-with-dockers:
     docker:
-      - image: 'alpine:3.7'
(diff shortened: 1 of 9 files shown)


brecke:
Bad idea because it would imply having docker calls inside a docker container and docker  inception is no good.

PR #1787, opened 2019-07-12
fix: make sure all previews are written to svg and html

### LINKED ISSUE #1786: Unable to see previews if it's a HTML file (not an SVG)
We're getting 404 trying to access content which has been previewed as an HTML file and not an SVG.

brecke:
Fixes #1786

---
This change is [Reviewable](https://reviewable.io/reviews/oaeproject/hilary/1787)


brecke:
This PR will create both an `svg` and a `html` preview so all files can be re-processed and the `svg` (which the 3akai-ux fronetnd asks for) is going to be there.

